PR #9, opened 2015-03-19
Reuse of SessionInputBuffer and SessionOutputBuffer internal buffers

srouthiau:
Commit https://github.com/apache/httpcore/commit/47654da6f027af3cbbd27600ba6df0911e529387

Reusing of internal buffer to limit memory usage. Performance improvements on our load tests.


ok2c:
What number in percent are we looking at?

srouthiau:
It was about 9 % of our total memory consumption. I don't have a test with
only httpcore and httpclient

httpcore usage on SessionInputBuffer was the 2nd on profiler hot spot
analysis.

ok2c:
Thread locals are really a double-edged sword and should be used sparingly in low level libraries. Can we provide a strategy interface similar to SSLBufferManagementStrategy in order to support multiple allocation strategies?

srouthiau:
This opt was done in the scope of one of my customer mission. I don't know
if he want to invest more time for this.

I'll ask him next time and let you know.

For the use of ThreadLocal, il agree it can lead to reuse large (too large)
buffers, that's why i have included a test to check if the buffer is not
too large to we can reuse it.

Stephane

2015-03-19 16:47 GMT+01:00 Oleg Kalnichevski <notifications@github.com>:

PR #10, opened 2015-06-03
Enable https protocol for TestHttpsAsyncTimeout, so timeout will be tested for real

hirthwork:
Without this test finishes immediately with IOException: SSL not supported, so not real handshake testing performed


ok2c:
Committed to SVN and 4.4.x branch 
http://svn.apache.org/r1683402

PR #12, opened 2015-08-25
Httpcore-406: more tests

bimargulies:
These are _really_ osgi tests.


ok2c:
Change-set committed to SVN trunk and 4.4.x branch.

PR #13, opened 2015-08-29
HTTPCORE-407: Add a ByteBufferEntity.

bimargulies:
HTTPCORE-407: Add a ByteBufferEntity class.


ok2c:
Patch committed to SVN trunk
http://svn.apache.org/r1700104

PR #17, opened 2015-11-09
HTTPCORE-412 trailers as Header[] and getExpectedTrailerNames

yaitskov:
getExpectedTrailerNames is introduced in HttpEntity for composing Trailer header.


ok2c:
DefaultHttpRequestWriter is the wrong place to add headers. This should be done by a protocol interceptor. But I can take from  there.

ok2c:
Patch with changes committed to SVN trunk
http://svn.apache.org/viewvc?rev=1713993&view=rev

PR #20, opened 2016-01-31
HTTPCORE-418: add constructor to HttpHost for hostname and scheme

HendricksonJS:
Add a new constructor that takes a hostname and port and uses the default port for that scheme.


garydgregory:
Patch applied:

Committed revision 1727859.

Please verify and close.

Thank you @HendricksonJS !

PR #21, opened 2016-02-05
Fix race condition in org.apache.http.nio.protocol.HttpAsyncRequestExecutor.endOfInput and org.apache.http.impl.nio.client.AbstractClientExchangeHandler.connectionAllocated(conn)

hirthwork:
Got pretty bad race condition when sending async requests.
Symptoms & Environment: We have target host which sometimes closes connections after requests processing but no `Connection: close` header was set. Some requests hangs, i.e. `BasicFuture` methods `.cancelled`, `.completed` or `.failed` methods are never called.

Analysis:
Let there be two threads:
- T1 is the async client reactor thread.
- T2 is the thread that initiates HTTP request.

T1: completes HTTP request and calls `connmgr.releaseConnection`
T2: initiates HTTP requests asks connmgr for connection and immediately gets connection that was just released by T1
T1: receives EOF on released connection and falls into `HttpAsyncRequestExecutor.endOfInput`. Connection has valid `HTTP_EXCHANGE_STATE`, but `HTTP_HANDLER` is null
T2: sets `HTTP_HANDLER`, requests output and checks `.isState`. Everything is OK here.
T1: goes to the end of `.endOfInput` and closes connection. `HTTP_HANDLER` is not null anymore, but is won't be notified about connection close, soon connection will be garbage collected and it will never be released and `HTTP_HANDLER.basicFuture` will never be notified.

Suggested fix (not sure if it is optimal, but it changed behaviour to be more redundant):
At the start of `.endOfInput` try to notify `HTTP_HANDLER`. If there is no one, set variable `handlerNotified` to false.
After connection close check `handlerNotified` and if it wasn't, then check if somebody has set `HTTP_HANDLER` externally.

### DIFF
--- a/httpcore-nio/src/main/java/org/apache/http/nio/protocol/HttpAsyncRequestExecutor.java
+++ b/httpcore-nio/src/main/java/org/apache/http/nio/protocol/HttpAsyncRequestExecutor.java
@@ -396,11 +396,11 @@ public class HttpAsyncRequestExecutor implements NHttpClientEventHandler {
         this.exceptionLogger.log(ex);
     }
 
-    private State getState(final NHttpConnection conn) {
+    private static State getState(final NHttpConnection conn) {
         return (State) conn.getContext().getAttribute(HTTP_EXCHANGE_STATE);
     }
 
-    private HttpAsyncClientExchangeHandler getHandler(final NHttpConnection conn) {
+    private static HttpAsyncClientExchangeHandler getHandler(final NHttpConnection conn) {
         return (HttpAsyncClientExchangeHandler) conn.getContext().getAttribute(HTTP_HANDLER);
     }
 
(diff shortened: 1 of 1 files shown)


ok2c:
Good catch. I'll review and commit the patch in the coming days.

ok2c:
Dmitry,
Curiously enough I had to fix a very similar race condition not that long ago. See 4c5947e4913c983b905d9d79db123a5b900a8b9a.

Would you be OK with a slightly different fix, which is essentially the same as yours but synchronizes critical section on the connection context?   
https://github.com/ok2c/httpcore/commit/38a2f7872b55019fbfad9a6989f1ea7a5cdd8801

hirthwork:
Yeah, I've seen 4c5947e4913c983b905d9d79db123a5b900a8b9a during race condition investigation, it was a very useful for me.
Your patch seems pretty valid for me, I'll be able to test it tomorrow.

ok2c:
Wonderful. I'll commit the fix to the ASF repository once I get your confirmation.

hirthwork:
After 1 hour of load testing no connection leaks detected. I suggest to consider ok2c@38a2f78 as fix for this race condition

ok2c:
Committed to SVN trunk and 4.4.x branch

PR #26, opened 2016-03-10
Code quality fix - Utility classes should not have public constructors.

faisal-hameed:
This pull request is focused on resolving occurrences of Sonar rule 
squid:S1118 - Utility classes should not have public constructors.
You can find more information about the issue here: 
https://dev.eclipse.org/sonar/rules/show/squid:S1118
 
Please let me know if you have any questions.

Faisal Hameed


garydgregory:
I think that if you want to have this kind of change, I would like it better if there was a code comment on the ctor that read something like "Do not allow utility this class to be instantiated."

faisal-hameed:
@garydgregory Done changes as you suggested.

garydgregory:
Thank you for the updated patch!

Committed revision 1734489.

Gary

PR #28, opened 2016-04-04
Support 64-bit chunk headers

hirthwork:
Some caching servers (like nginx) doesn't send Content-Length for cached responses from upstream which returned chunked response. The only way to ensure response completeness is to use chunked encoding.
Unfortunately, nginx sends whole response in single chunk, so chunk header will contain file size, which can be greater than Integer.MAX_VALUE (however, still can be processed by ZeroCopyConsumer).

Proposed patch contains changes for huge chunks processing. New unit-tests included.


ok2c:
Patch committed to SVN trunk and 4.4.x branch. Please review and close

PR #31, opened 2016-10-06
(HTTPCLIENT-1759) Contribute new tests that use the in-process server.

lewijw:
Please see HTTPCLIENT-1759 for details.

Also, I copied URLEncodedUtils from httpclient5 to org.apache.hc.core5.util.
Also, I added an additional shutdown method to ClassicTestServer so the grace period can be set.


ok2c:
Committed to SVN trunk. Please review and close.

PR #32, opened 2016-12-21
HTTPCORE-439: Basicheader override of equals and hashcode

lewijw:
When removing headers from a HeaderGroup, the header is only removed if the header you are trying to remove is an object in the header list. It seems it would be better if headers were removed that matched the name and value of one of the objects in the list. To accomplish this, I have overridden the equals and hashCode methods in BasicHeader.


ok2c:
Committed to SVN trunk with some changes
http://svn.apache.org/viewvc?rev=1775588&view=rev

Please review and close.

lewijw:
Thanks for making the changes and expediting this to get it in the alpha2 release.  I got rid of some redundant test code.  I am also assuming these headers are equal (the sensitivity is of no concern with equality):
        final Header header = new BasicHeader("name", "value", true);
        final Header header2 = new BasicHeader("name", "value", false);

When is the alpha2 release?

ok2c:
Committed to SVN trunk
URL: http://svn.apache.org/viewvc?rev=1775806&view=rev

Please review and close.

PS: The 5.0-alpha2 release vote is ongoing. However due to the time of the year I cannot rule out its failure due to insufficient binding votes. If I have to cut another RC your changes will get picked up.

lewijw:
Thanks!

PR #37, opened 2017-02-07
[HTTPCLIENT-1810] URIBuilder.toString() always adds '/' to path, even with query parameters 

Coderec:
When the path is null or '', URIBuilder.normalizePath() should return '' rather than '/'.


ok2c:
Committed to SVN trunk. Please review and close.

seancarroll:
There still appear to be an issue with normalizePath starting with 4.5.3 which adds a leading '/'. If you do normalizePath("https://www.google.com/"); the result is /https://www.google.com/

PR #40, opened 2017-03-30
add a provider parameter in SSLContextBuilder

lujianbo:
[HTTPCORE-450]
add a provider parameter in SSLContextBuilder to support third party provider

https://issues.apache.org/jira/browse/HTTPCORE-450

### DIFF
--- a/httpcore/src/main/java/org/apache/http/ssl/SSLContextBuilder.java
+++ b/httpcore/src/main/java/org/apache/http/ssl/SSLContextBuilder.java
@@ -33,7 +33,16 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.net.Socket;
 import java.net.URL;
-import java.security.*;
+import java.security.KeyManagementException;
+import java.security.KeyStore;
+import java.security.KeyStoreException;
+import java.security.NoSuchAlgorithmException;
+import java.security.NoSuchProviderException;
+import java.security.Principal;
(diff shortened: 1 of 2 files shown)


httpcore/src/main/java/org/apache/http/ssl/SSLContextBuilder.java
```diff
@@ -272,6 +274,20 @@ public SSLContext build() throws NoSuchAlgorithmException, KeyManagementExceptio
         return sslcontext;
     }
 
+    public SSLContext buildWithProvider(final Provider provider) throws NoSuchAlgorithmException, KeyManagementException {
```
ok2c: Do it really take a special `#build` method? Can't we have a `#useProvider` setter similar to `#useProtocol` ?

lujianbo: buildWithProvider(final String provider) will throw a new NoSuchProviderException.

in order to be compatible with the old interface so add a new interface

ok2c: Fair point. However I am not sure this change should go into 4.4. In 5.0 we can change APIs as we deem fit.

lujianbo: i agree

PR #41, opened 2017-03-31
add a provider parameter in SSLContextBuilder In 5.0 branch

lujianbo:
[HTTPCORE-450]
add a provider parameter in SSLContextBuilder In 5.0 branch
referrer:https://issues.apache.org/jira/browse/HTTPCORE-450

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/ssl/SSLContextBuilder.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/ssl/SSLContextBuilder.java
@@ -37,8 +37,10 @@ import java.security.KeyManagementException;
 import java.security.KeyStore;
 import java.security.KeyStoreException;
 import java.security.NoSuchAlgorithmException;
+import java.security.NoSuchProviderException;
 import java.security.Principal;
 import java.security.PrivateKey;
+import java.security.Provider;
 import java.security.SecureRandom;
 import java.security.UnrecoverableKeyException;
 import java.security.cert.CertificateException;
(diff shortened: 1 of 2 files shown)


lujianbo:
@garydgregory  no, when Provider Object and provider name have been set in the same time，it will choose the Provider object first。

garydgregory:
I can see that. My point is that it is confusing to allow a user to set both and have one used and not the other when this is not even documented in a Javadoc comment. Do you really need both? @ok2c, any thoughts?

lujianbo:
there are some documents about SSLContext.

http://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLContext.html#getInstance-java.lang.String-java.security.Provider-

http://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLContext.html#getInstance-java.lang.String-java.lang.String-

ok2c:
@lujianbo Builders with mutually exclusive settings have been causing a lot of grief in HC 4.x. We are trying to not repeat the same mistake in HC 5.0. I think `#useProvider(String)` could be dropped. One can always resolve providers by their name with [Security#getProvider](useProviderl) call.

lujianbo:
@ok2c  ok，i wil fix it and crate a new pull request.

ok2c:
You can push to the same branch. No need to create a new PR

lujianbo:
@garydgregory @ok2c may i pull this change to 4.4.7 to support provider.

garydgregory:
Based on our current versioning policy and recent discussions about other features, this would not be appropriate in a bug fix version like 4.4.7 but should go in as a new API/feature in a 4.5.x line. @ok2c ?

ok2c:
We have already repeatedly violated this policy in 4.4.x, so we might as well do it again, as long as it is just a setter method.

garydgregory:
@lujianbo so it sounds like "go for it". I'd suggest you port the trunk code to the 4.4 branch in a new PR.

lujianbo:
ok

PR #46, opened 2017-08-05
SSLBufferManagement can be overridden for custom buffer pooling

cakofony:
Some applications maintain ByteBuffer pools already, reusing
already pooled buffers with the client can reduce memory churn.

Additionally, using native crypto JSSE implementations (such as
wildfly-openssl or conscrypt/boringssl) requires the use of direct
ByteBuffers.


ok2c:
Please review and close the PR.

PR #47, opened 2017-08-07
Fix hang when using AbstractClassicEntityProducer

cakofony:
When AsyncDataProducer.available returns zero, the caller may
not execute the produce method again until
DataStreamChannel.requestOutput has been invoked.

This approach sets available to Integer.MAX_VALUE to avoid this
case. It would be more efficient to call requestOutput after
writes, but I'm not certain if invoking the method unnecessarily
would affect performance, and the MAX_VALUE approach is used by
other producers.

I was able to reproduce this issue in a test by sending a dummy 4gb
stream using AbstractClassicEntityProducer. Overriding available with
a non-zero value stopped the hangs.
Again, it may be more efficient to track `available()` invocations and
call `requestOutput()` from a subtype of ContentOutputStream, but I
haven't dug in quite that far yet :-)

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/http/nio/entity/AbstractClassicEntityProducer.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/http/nio/entity/AbstractClassicEntityProducer.java
@@ -62,7 +62,7 @@ public abstract class AbstractClassicEntityProducer implements AsyncEntityProduc
 
     @Override
     public final int available() {
-        return buffer.length();
+        return Integer.MAX_VALUE;
     }
 
     @Override


ok2c:
@cakofony Always returning Integer.MAX_VALUE from `AsyncDataProducer#available` does not sound right me. Simple producers do that if they are self-contained and can always produce more data if the underlying channel is ready. This is not the case with complex producers such as `AsyncDataProducer` that may not always have data to write to the channel. By returning Integer.MAX_VALUE from `AsyncDataProducer#available` one can cause the I/O reactor spin busily and produce no data. 

We need to think of a better fix. How reliably can you reproduce the issue and would it be possible to make a junit?

ok2c:
@cakofony Did you have any luck reproducing the issue with an unit / integration test?

cakofony:
Apologies, I've been busy, I will take a crack at it this week!

PR #50, opened 2017-08-29
setting time limit on processPending methods

xiaohu-zhang:
setting time limit to avoid the SingleCoreIOReactor's time be exhausted
and then  block the remaining work to do.

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/reactor/SingleCoreIOReactor.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/reactor/SingleCoreIOReactor.java
@@ -178,7 +178,7 @@ class SingleCoreIOReactor extends AbstractSingleCoreIOReactor implements Connect
 
     private void processPendingChannels() throws IOException {
         SocketChannel socketChannel;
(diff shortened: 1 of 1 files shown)


ok2c:
Is there any particular reason for choosing the limit of 100000? Is this change related to HTTPCORE-484 in any way?

xiaohu-zhang:
It is the change related to HTTPCORE-486.100000 is an arbitrary value.In my test ,loop 100000 times
will take less than 3ms usually. 100000 is also the value netty used as the similar logic .

xiaohu-zhang:
this change is not related to HTTPCORE-484

ok2c:
Committed with minor changes. Please review and close the PR.

xiaohu-zhang:
thank you，the commit is ok to me.

yinwoods:
> It is the change related to HTTPCORE-486.100000 is an arbitrary value.In my test ,loop 100000 times will take less than 3ms usually. 100000 is also the value netty used as the similar logic .

It's said to use 100K as limit value, but actually MAX_CHANNEL_REQUESTS = 10000. Is this a typo bug？I have found when send too many requests will cause FullGC, even OOM

PR #52, opened 2017-09-27
HTTPCORE-491 make basicasyncresponseconsumer paranoid

mheemskerk:
`BasicAsyncRequestConsumer` and `BasicAsyncResponseConsumer` used to
blindly pre-allocate a buffer of `Content-Length`. If the request / response
is crafted to have a very large `Content-Length` header, this can result in
`OutOfMemoryError`s.

This limits the initial buffer size to 256kb, while still allowing the
buffer to grow if a larger message is read.


ok2c:
Committed to 4.4.x branch. Please review and close.

PR #55, opened 2017-12-14
Fix request splitting

werehuman:
If user has access to any header value, he can add any additional malicious header, like `Host`, `X-Forwarded-Host` or even make another HTTP request.

http://projects.webappsec.org/w/page/13246929/HTTP%20Request%20Splitting


ok2c:
@werehuman Committed to 4.4.x and master. Please review and close this PR.

PR #57, opened 2018-01-19
Workaround for misbehaved servers that return HTTP 204 responses with a content

agherardi:
I have encountered a server that incorrectly returns a content on a HTTP 204 response. The server also returns a Content-length header with the response.

httpclient assumes that HTTP 204 responses have no content, so it doesn't consume the content.

If the connection is reused, this causes httpclient to hang when looking for the status line of the next response.

This patch attempts to prevent the issue by allowing HTTP 204 responses with Content-length or Transfer-encoding: Chunked as responses to have a content.

### DIFF
--- a/httpcore/src/main/java/org/apache/http/impl/DefaultConnectionReuseStrategy.java
+++ b/httpcore/src/main/java/org/apache/http/impl/DefaultConnectionReuseStrategy.java
@@ -80,17 +80,26 @@ public class DefaultConnectionReuseStrategy implements ConnectionReuseStrategy {
         Args.notNull(response, "HTTP response");
         Args.notNull(context, "HTTP context");
 
-        // If a HTTP 204 No Content response contains a Content-length or Transfer-encoding:Chunked header,
+        // If a HTTP 204 No Content response contains a Content-length with value > 0 or Transfer-encoding:Chunked header,
         // don't reuse the connection. This is to avoid getting out-of-sync if a misbehaved HTTP server
         // returns content as part of a HTTP 204 response.
         if (response.getStatusLine().getStatusCode() == HttpStatus.SC_NO_CONTENT) {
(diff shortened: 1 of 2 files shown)


reschke:
That sounds like a bad idea. It would treat a 204 *with* Content-Length but without payload incorrectly, right?

agherardi:
> It would treat a 204 with Content-Length but without payload incorrectly, right?

Correct. However, the presence of a Content-Length is more indicative that a payload exists than the status code 204 being used incorrectly.

reschke:
Both are spec violations. Maybe this should be treated as fatal error.

ok2c:
I fully agree with @reschke. Being 'clever' about such severe protocol violations is not good idea. However I would happily commit a patch failing 204 responses with `Content-Length` or `Transfer-Encoding` headers

agherardi:
Thanks for your feedback.

Failing the response should work for me. However, in that case httpclient should also close the connection - if it goes back to the pool it'll cause troubles since client and server are out-of-sync. Do you have any suggestions on how to do that?

agherardi:
One possibility would be to modify DefaultClientConnectionReuseStrategy.keepAlive() as follows: If the response is a HTTP 204 and contains either content-length or transfer-encoding, return FALSE - i.e., do NOT reuse the connection.

Sounds acceptable? Any better alternatives?

ok2c:
@agherardi An alternative would be to throw a `ProtocolException` from a response interceptor, however the solution you have proposed looks better to me.

agherardi:
Closed and replaced by https://github.com/apache/httpcomponents-client/pull/95

agherardi:
Changed approach and modified HttpCore DefaultConnectionReuseStrategy as suggested. Added unit-tests.

ok2c:
@agherardi Committed to master and 4.4.x with some minor tweaks. Please review

agherardi:
Looks good. Thank you!

PR #58, opened 2018-02-12
ExpandableBuffer breaks on large response

imoldovan-intacct:
Ticket: https://issues.apache.org/jira/browse/HTTPCORE-513


ok2c:
@imoldovan-intacct Your code makes use of Java 1.8 specific features which makes the change-set incompatible with the current versions of HttpCore. HttpCore 4.4 still requires Java 1.6 and HttpCore 5.0 requires Java 1.7. It is not a big deal for me to fix it it would help if you could amend your patch. It would also help if you could fix style-check violations.

imoldovan-intacct:
Will do. I only solved my own problem, which is selfish :) .

imoldovan-intacct:
This closes #58.

PR #59, opened 2018-02-15
Avoid endless loop

imoldovan-intacct:
Of course, I did not think this through...
This pull request is related to https://github.com/apache/httpcomponents-core/pull/58 .

We might get into an endless loop for sizes which are in (Integer.MAX_VALUE - headRoom, Integer.MAX_VALUE] .

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/http/impl/nio/ExpandableBuffer.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/http/impl/nio/ExpandableBuffer.java
@@ -110,16 +110,18 @@ public class ExpandableBuffer {
     protected void expand() {
         int newcapacity = (this.buffer.capacity() + 1) << 1;
         if (newcapacity < 0) {
-            int headRoom = Integer.max(Long.BYTES, 8);
-            // Reason: In GC the size of objects is passed as int (2 bytes). 
-            // Then, the header size of the objects is added to the size. 
-            // Long has the longest header available. Object header seems to be linked to it. 
-            // Details: I added a minimum of 8 just to be safe and because 8 is used in 
+            final int vmBytes = Long.SIZE >> 3;
+            final int javaBytes = 8; // this is to be checked when the JVM version changes
+            @SuppressWarnings("unused") // we really need the 8 if we're going to make this foolproof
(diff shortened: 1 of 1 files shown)


httpcore5/src/main/java/org/apache/hc/core5/http/impl/nio/ExpandableBuffer.java
```diff
@@ -123,6 +123,10 @@ protected void expand() {
             // WARNING: This code assumes you are providing enough heap room with -Xmx.
             // source of inspiration: https://bugs.openjdk.java.net/browse/JDK-8059914
             newcapacity = Integer.MAX_VALUE - headRoom;
+            if (newcapacity == this.buffer.capacity()) {
```
ok2c: @imoldovan-intacct  why not `newcapacity >= this.buffer.capacity()` ?

ok2c: @imoldovan-intacct you might also want to rebase your branch off the upstream master.

imoldovan-intacct: I did that.
I also did a beginner mistake :) . I'll try not to commit with too much coffee on board anymore.

imoldovan-intacct: There is no way for this.buffer.capacity() to exceed Integer.MAX_VALUE - headRoom.
When we get to the point that they are equal => we already have touched maximum buffer capacity allowed once (MAX_VALUE - 8 in our case) and we are going inside an endless loop.

ok2c: bq. There is no way for this.buffer.capacity() to exceed Integer.MAX_VALUE - headRoom
@imoldovan-intacct  Probably, but still looks a little safer

ok2c:
@imoldovan-intacct I would rather use `BufferOverflowException` instead of `MessageConstraintException`

imoldovan-intacct:
Thanks for the suggestions. It's all good now.
I do have some improvements in mind for the buffer allocation, but I preferred to stick to the burning issue for now.

PR #64, opened 2018-05-18
Add remote address to connect exception

jasontedor:
Today when timing out a connection the exception message is not populated. This is not helpful for end users. This commit adds an exception message that indicates the cause of the connect exception is a timeout to the remote address.


garydgregory:
Sounds reasonable to me. @ok2c ?

ok2c:
@jasontedor @garydgregory Works for me.

garydgregory:
I committed this PR with some minor changes to the branch 4.4.x. Please verify and close.

PR #65, opened 2018-05-31
Add Automatic-Module-Name in manifest

varunnvs92:
Add Automatic-Module-Name in manifest so Java9 modular applications can depend on this library.

It is recommended for java libraries in public repositories like Maven Central to add Automatic-Module-Name manifest entry. This allows the consumers of the library to move to Java 9 modularization and depend on the Automatic-Module-Name value instead of the name derived from the Jar file. This way, your consumers won't need to make changes when your library is eventually converted to a Java 9 module.

Adding Automatic-Module-Name doesn't affect any code and will simply make a name reservation until an actual module-info.java will be added eventually. More details can be found in this [article](https://dzone.com/articles/automatic-module-name-calling-all-java-library-maintainers).

[This article](http://blog.joda.org/2017/04/java-se-9-jpms-module-naming.html) has best naming practices for JPMS modules. But I could not go with ```org.apache.http``` (reverse dns package name) as all your modules use the same base package name. I think ```org.apache.httpcomponents.core``` seems more targeted. Let me know if you have better naming suggestions. 

[AWS Java SDK v2](https://github.com/aws/aws-sdk-java-v2) has a dependency on httpclient library which transitively depends on this libarary. We are adding module-info files to our library. But we need the Automatic-Module-Name in both this library and [httpcomponents-client](https://github.com/apache/httpcomponents-client) library before we can move to JPMS. Please prioritize this issue and we request a release with this change by mid June. If this is approved, I will send a PR to httpcomponents-client repo.

I checked effective pom to verify that this change will still bring all other manifestEntries from the parent pom.


ok2c:
@varunnvs92 Is there a reason for adding `Automatic-Module-Name` to HttpCore main module and not to any other modules?

varunnvs92:
I started with the main module as our library depends on it and this PR is meant to start the discussion about becoming JPMS compatible. I can update the PR to add it to other modules too. Let me know if the following module names are acceptable:

httpcore: ```org.apache.hc.http.core```
httpcore-nio: ```org.apache.hc.http.nio```
httpcore-ab: ```org.apache.hc.http.benchmark```

I have added 'hc' in all the names, so if you have other libraries that have core, nio modules they can evolve separately. Let me know if you have better naming suggestions. I am not sure if we need Automatic module name for osgi module.

varunnvs92:
Updated the PR with the suggested module names

varunnvs92:
When do you plan to release a new version to Maven that has this change?

ok2c:
@varunnvs92 Hopefully by the end of the month.

PR #68, opened 2018-07-10
HTTPCORE-532

xiaohu-zhang:
IOReactor thread can be lazy started

### DIFF
--- a/httpcore5-testing/src/main/java/org/apache/hc/core5/testing/nio/IOReactorExecutor.java
+++ b/httpcore5-testing/src/main/java/org/apache/hc/core5/testing/nio/IOReactorExecutor.java
@@ -85,7 +85,10 @@ abstract class IOReactorExecutor<T extends IOReactorService> implements AutoClos
 
                 }))) {
             if (status.compareAndSet(Status.READY, Status.RUNNING)) {
-                ioReactorRef.get().start();
+                if(this instanceof AsyncServer){
+                    //start DefaultListeningIOReactor
+                    ioReactorRef.get().start(0);
+                }
             }
         } else {
             throw new IllegalStateException("I/O reactor has already been started");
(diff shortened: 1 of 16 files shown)


xiaohu-zhang:
there is somthing wrong with the code ,let me correct it first

httpcore5-testing/src/main/java/org/apache/hc/core5/testing/nio/IOReactorExecutor.java
```diff
@@ -85,7 +85,10 @@ public void execute(final IOSession session) {
 
                 }))) {
             if (status.compareAndSet(Status.READY, Status.RUNNING)) {
-                ioReactorRef.get().start();
+                if(this instanceof AsyncServer){
```
ok2c: @xiaohu-zhang This is terrible. Even for test classes we should not be doing things like that.

httpcore5/src/main/java/org/apache/hc/core5/http/impl/bootstrap/AsyncServer.java
```diff
@@ -73,8 +73,12 @@ public AsyncServer(
     }
 
     @Override
-    public void start() {
-        ioReactor.start();
+    public void start(final int i) {
+        throw new RuntimeException("can not call start(i) method with AsyncServer class,maybe call serverStart method?");
```
ok2c: @xiaohu-zhang I cannot commit changes like that.

xiaohu-zhang:
when use few connections,for example ,use always one connection in connection pool each time(), or no http connection used when project started.framework can use less thread ,that will reduce the thread number and load in computer.

ok2c:
@xiaohu-zhang If you anticipate few connections and want fewer threads you really _should_ configure fewer workers. Overall, I think this change-set causes more problems than it solves.

xiaohu-zhang:
what if can not anticipate ?lazy init is little useful in  performance。Also the code can be modified do not use instanceof .after all , it depends on you.

ok2c:
@xiaohu-zhang Why? The I/O reactor distributes sessions across all available workers equally. If you have n worker threads, all of them will get started after n connections.

xiaohu-zhang:
yes.it has a little useful only when connections less than worker thread,and is not anticipate by user before.
just tiny improvement.

PR #70, opened 2018-07-15
Add HttpContext to AsyncPushConsumer#consumePromise

ato:
This is to enable the push consumer to determine which request the server sent the push in reply to when there are multiple multiplexed requests active.

There may be a more appropriate name than 'original context'. I also considered 'source context' and 'pushed from context' but wasn't totally happy with any of them. I'm also unsure if there's a better way to access the context than making it package-private like this.

This is the core part of solving the problem I described in "Associating a server push with the original client request" [thread] on the users list.

The client part is to change AsyncPushConsumer to receive the push's HttpContext. It currently only has access to the push's HttpRequest and HttpResponse. The push consumer could then do something like:

    context.getOriginalContext().getRequest();

In my 'saving to a directory' example when I make the request add a custom context attribute for the directory I want to save to and then access it in the push consumer like:

     context.getOriginalContext().getAttribute("myapp.request-specific-output-directory");

[thread]: http://mail-archives.apache.org/mod_mbox/hc-httpclient-users/201807.mbox/%3C1531636562.3812678.1441097240.7E58FF13%40webmail.messagingengine.com%3E

### DIFF
--- a/httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/ClientHttp2StreamMultiplexer.java
+++ b/httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/ClientHttp2StreamMultiplexer.java
@@ -83,10 +83,15 @@ public class ClientHttp2StreamMultiplexer extends AbstractHttp2StreamMultiplexer
     Http2StreamHandler createRemotelyInitiatedStream(
             final Http2StreamChannel channel,
             final HttpProcessor httpProcessor,
-            final BasicHttpConnectionMetrics connMetrics) throws IOException {
+            final BasicHttpConnectionMetrics connMetrics,
+            final Http2StreamHandler originalHandler) throws IOException {
         final HttpCoreContext context = HttpCoreContext.create();
         context.setAttribute(HttpCoreContext.SSL_SESSION, getSSLSession());
         context.setAttribute(HttpCoreContext.CONNECTION_ENDPOINT, getEndpointDetails());
+        if (originalHandler != null && originalHandler instanceof ClientHttp2StreamHandler) {
+            final HttpCoreContext originalContext = ((ClientHttp2StreamHandler) originalHandler).context;
+            context.setAttribute(HttpCoreContext.ORIGINAL_CONTEXT, originalContext);
+        }
         return new ClientPushHttp2StreamHandler(channel, httpProcessor, connMetrics, pushHandlerFactory, context);
     }
 
--- a/httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/ClientHttp2StreamMultiplexer.java
+++ b/httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/ClientHttp2StreamMultiplexer.java
@@ -32,7 +32,6 @@ import org.apache.hc.core5.http.config.CharCodingConfig;
 import org.apache.hc.core5.http.impl.BasicHttpConnectionMetrics;
 import org.apache.hc.core5.http.nio.AsyncPushConsumer;
 import org.apache.hc.core5.http.nio.HandlerFactory;
-import org.apache.hc.core5.http.protocol.HttpContext;
 import org.apache.hc.core5.http.protocol.HttpCoreContext;
 import org.apache.hc.core5.http.protocol.HttpProcessor;
 import org.apache.hc.core5.http2.config.H2Config;
--- a/httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/ClientHttp2StreamMultiplexer.java
+++ b/httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/ClientHttp2StreamMultiplexer.java
@@ -88,9 +88,8 @@ public class ClientHttp2StreamMultiplexer extends AbstractHttp2StreamMultiplexer
         final HttpCoreContext context = HttpCoreContext.create();
         context.setAttribute(HttpCoreContext.SSL_SESSION, getSSLSession());
         context.setAttribute(HttpCoreContext.CONNECTION_ENDPOINT, getEndpointDetails());
-        if (originalHandler != null && originalHandler instanceof ClientHttp2StreamHandler) {
-            final HttpCoreContext originalContext = ((ClientHttp2StreamHandler) originalHandler).context;
-            context.setAttribute(HttpCoreContext.ORIGINAL_CONTEXT, originalContext);
+        if (originalHandler != null) {
+            context.setAttribute(HttpCoreContext.ORIGINAL_CONTEXT, originalHandler.getContext());
         }
         return new ClientPushHttp2StreamHandler(channel, httpProcessor, connMetrics, pushHandlerFactory, context);
     }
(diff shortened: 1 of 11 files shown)


httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/ClientHttp2StreamMultiplexer.java
```diff
@@ -83,10 +83,15 @@ public ClientHttp2StreamMultiplexer(
     Http2StreamHandler createRemotelyInitiatedStream(
             final Http2StreamChannel channel,
             final HttpProcessor httpProcessor,
-            final BasicHttpConnectionMetrics connMetrics) throws IOException {
+            final BasicHttpConnectionMetrics connMetrics,
+            final Http2StreamHandler originalHandler) throws IOException {
         final HttpCoreContext context = HttpCoreContext.create();
         context.setAttribute(HttpCoreContext.SSL_SESSION, getSSLSession());
         context.setAttribute(HttpCoreContext.CONNECTION_ENDPOINT, getEndpointDetails());
+        if (originalHandler != null && originalHandler instanceof ClientHttp2StreamHandler) {
```
ok2c: @ato I agree in principle but can this cast be avoided?

httpcore5/src/main/java/org/apache/hc/core5/http/protocol/HttpCoreContext.java
```diff
@@ -155,4 +162,10 @@ public HttpResponse getResponse() {
         return getAttribute(HTTP_RESPONSE, HttpResponse.class);
     }
 
+    /**
+     * @since 5.0
+     */
+    public HttpContext getOriginalContext() {
```
ok2c: @ato Do we need to change `HttpCoreContext` at all? This seems like a very special case specific to HTTP/2 push.

ato: Good point. I'll remove the getter.

Should the attribute name constant go as well? Do you have a suggestion for a more appropriate class to put the constant in?

httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/ClientHttp2StreamMultiplexer.java
ato: Oops. I'll add a `getContext()` method to Http2StreamHandler. I mistakenly thought it was a public interface and was trying to avoid a breaking change. I should have read the code more carefully, it's package private.

httpcore5/src/main/java/org/apache/hc/core5/http/nio/AsyncPushConsumer.java
```diff
@@ -40,7 +41,7 @@
  */
 public interface AsyncPushConsumer extends AsyncDataConsumer {
 
-    void consumePromise(HttpRequest promise, HttpResponse response, EntityDetails entityDetails) throws HttpException, IOException;
+    void consumePromise(HttpRequest promise, HttpResponse response, EntityDetails entityDetails, HttpContext context) throws HttpException, IOException;
```
ato: Note this one definitely is a public API change. Is changing the interface like this acceptable?

ok2c: @ato HC 5.0 is still BETA. We are at liberty to make small changes if it helps keep APIs clean

ato:
@ok2c Pull request updated to address review comments.

I'll send a pull request against httpcomponents-client for the corresponding interface changes shortly.

httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/ClientHttp2StreamMultiplexer.java
```diff
@@ -88,9 +88,8 @@ Http2StreamHandler createRemotelyInitiatedStream(
         final HttpCoreContext context = HttpCoreContext.create();
         context.setAttribute(HttpCoreContext.SSL_SESSION, getSSLSession());
         context.setAttribute(HttpCoreContext.CONNECTION_ENDPOINT, getEndpointDetails());
-        if (originalHandler != null && originalHandler instanceof ClientHttp2StreamHandler) {
```
ok2c: @ato What if passed the original request message (or its request URI) instead of linking the contexts?

ato:
> @ato What if we passed the original request message (or its request URI) instead of linking the contexts?

For my use case the request message is a means to an end. I want to associate an id or some application data with the request and then get it back again so that I know what to do with that particular push when I receive it.

The URI alone wouldn't work as I might be making multiple requests to same the URI such as multiple POSTs with a different body.

Technically I guess the identity of the request object would work as an id but it would be somewhat indirect. I'd probably keep an IdentityHashMap<HttpRequest,MyData> and then pass my per-request application data into the context:

    requestMap.put(request, context.getAttribute("mydata"));
    client.execute(request, consumer, context, callback); 

Then in push consumer I could get at it like:

    var mydata = globalRequestMap.get(originalRequest);

Or I guess maybe subclass the request and add a field for my data?

Here's another alternative. What if rather than registering the AsyncPushConsumers on the HttpClient you could pass a AsyncPushConsumer when making the request:

    var pushConsumer = new MyPushConsumer(mydata);
    client.execute(request, responseConsumer, pushConsume);

That would totally solve my problem in a fairly clean way without exposing the context or changing the AsyncPushConsumer interface.

I noticed that's how the JDK11 HttpClient solves this problem:

https://download.java.net/java/early_access/jdk11/docs/api/java.net.http/java/net/http/HttpClient.html#sendAsync(java.net.http.HttpRequest,java.net.http.HttpResponse.BodyHandler,java.net.http.HttpResponse.PushPromiseHandler)

ato:
Edit: Removed the context nonsense from the IdentityHashMap example. I hit the comment button too soon. :-(

ok2c:
> Here's another alternative. What if rather than registering the AsyncPushConsumers on the HttpClient you could pass a AsyncPushConsumer when making the request:

@ato I had a nagging feeling when developing HTTP/2 push functionality that would become necessary at some point. This is indeed a better approach.

Can I suggest the following? Could you please reduce the scope of this change-set to adding `HttpContext` parameter to `AsyncPushConsumer#consumePromise` without any `HttpContext` chaining stuff? As the next step we can work on extending `HttpAsyncRequester` in core and `HttpAsyncClient` in client a method that can pass `AsyncPushConsumer` or `HttpRequestMapper` to the request execution handler.

ok2c:
@ato Please let me know if you intend to continue working on this change. I want to avoid duplicating your efforts.

ato:
@ok2c Sorry, some other issues came up and I didn't get a chance to progress this. I've removed the context chaining stuff.

Yes, I still want this. It sounds like you might want to make some progress on this yourself though. If so, I certainly won't complain. ;-) Otherwise I'll have a go at the client changes tomorrow. I'm still quite distracted today unfortunately.

ato:
Just realised getContext() can go too. One moment.

ok2c:
@ato Committed as 4ddfdbf6504fd1924373bab1c2a5b7b2879183e2. Please review and close.

ato:
Looks good. Thanks.

ok2c:
> Yes, I still want this. It sounds like you might want to make some progress on this yourself though. 

Not necessarily. As long as someone is looking into it and can happily step aside and work on something else.

ok2c:
@ato These are the API changes I would like to propose 45c537b84aae7ca7043d1591f74e9e203532562e

ato:
@ok2x That looks perfect to me.

If the user calls client. execute() with a push consumer it would override any registered via client.register() right? I don't think it'd make any sense to call multiple consumers.

ok2c:
> If the user calls client. execute() with a push consumer it would override any registered via client.register() right? 

Yes, it would.

PR #71, opened 2018-07-19
Implement proposed AsyncPushConsumer execution API 

ato:
@ok2c I tried to implement the proposed API. Some comments below.

### DIFF
--- a/httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/AbstractHttp2StreamMultiplexer.java
+++ b/httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/AbstractHttp2StreamMultiplexer.java
@@ -61,6 +61,7 @@ import org.apache.hc.core5.http.impl.BasicEndpointDetails;
 import org.apache.hc.core5.http.impl.BasicHttpConnectionMetrics;
 import org.apache.hc.core5.http.impl.CharCodingSupport;
 import org.apache.hc.core5.http.nio.AsyncClientExchangeHandler;
+import org.apache.hc.core5.http.nio.AsyncPushConsumer;
 import org.apache.hc.core5.http.nio.AsyncPushProducer;
 import org.apache.hc.core5.http.nio.command.ExecutionCommand;
 import org.apache.hc.core5.http.nio.command.ShutdownCommand;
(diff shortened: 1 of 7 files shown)


httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/ClientPushHttp2StreamHandler.java
```diff
@@ -99,16 +101,18 @@ public void consumePromise(final List<Header> headers) throws HttpException, IOE
...
-                throw new H2StreamResetException(H2Error.REFUSED_STREAM, "Stream refused");
+            if (exchangeHandler == null) {
+                final AsyncPushConsumer handler;
+                try {
+                    handler = pushHandlerFactory != null ? pushHandlerFactory.create(request, context) : null;
+                } catch (final ProtocolException ex) {
+                    throw new H2StreamResetException(H2Error.PROTOCOL_ERROR, ex.getMessage());
+                }
+                if (handler == null) {
+                    throw new H2StreamResetException(H2Error.REFUSED_STREAM, "Stream refused");
```
ato: One possible issue is the user cannot refuse a stream if they supply a push consumer to execute(). I see now that one of the purposes of the PushHandlerFactory is to allow early refusal.

httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/AbstractHttp2StreamMultiplexer.java
```diff
@@ -610,7 +612,7 @@ private void processPendingCommands() throws IOException, HttpException {
                         connMetrics,
                         exchangeHandler,
                         context);
-                final Http2Stream stream = new Http2Stream(channel, streamHandler, false);
+                final Http2Stream stream = new Http2Stream(channel, streamHandler, pushConsumer, false);
```
ato: We must stash the push consumer on some object that's attached to the original stream so we can later get at it when creating the new promised stream. I could see at least three ways of doing that:
- on the stream directly (as done here)
- on the stream handler
- adding an 'acceptPromise' method to the exchange handler.

I must had admit I was tempted to modify the exchange handler. But it didn't fit with your API proposal and I'm not confident enough with the code base to understand the design consequences.

ok2c: @ato Feel free to do what you deem right / proper. My proposal was just a quick sketch. Feel free to disregard it.

ato: Thanks. In that case I will close this for now and try the other approach.

ok2c:
@ato Are you still interested in contributing this feature or shall I take over?

ato:
@ok2c, I attempted a second try at this but struggled with the design. I won't be able to get back to this for some time now due to other work priorities that have come up. So if you would like to take over by all means do so.

ok2c:
@ato b66198e2de091195232eea96ea3593d2383bcab7#diff-253296d841a758a2fae53c7c42ebb670 does not feel very elegant but it gets it done. Could you please have a look if that is what you want? I am open to suggestions on adding convenience APIs

ok2c:
@ato Committed to master as 36ef4aa747e6a9a0b00cc93eb3c5482f62e6ba8b

PR #72, opened 2018-08-06
Fix HTTPCORE-536

clevertension:
Fix HTTPCORE-536

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/http/impl/bootstrap/AsyncRequester.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/http/impl/bootstrap/AsyncRequester.java
@@ -27,12 +27,6 @@
 
 package org.apache.hc.core5.http.impl.bootstrap;
 
-import java.io.IOException;
-import java.net.InetSocketAddress;
-import java.net.SocketAddress;
-import java.util.List;
-import java.util.concurrent.Future;
-
 import org.apache.hc.core5.concurrent.DefaultThreadFactory;
 import org.apache.hc.core5.concurrent.FutureCallback;
 import org.apache.hc.core5.function.Callback;
@@ -41,7 +35,6 @@ import org.apache.hc.core5.function.Resolver;
 import org.apache.hc.core5.http.HttpHost;
 import org.apache.hc.core5.http.impl.DefaultAddressResolver;
 import org.apache.hc.core5.io.CloseMode;
-import org.apache.hc.core5.net.NamedEndpoint;
 import org.apache.hc.core5.reactor.ConnectionInitiator;
 import org.apache.hc.core5.reactor.DefaultConnectingIOReactor;
 import org.apache.hc.core5.reactor.ExceptionEvent;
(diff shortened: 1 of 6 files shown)


httpcore5/src/main/java/org/apache/hc/core5/http/impl/bootstrap/AsyncRequester.java
```diff
@@ -27,12 +27,6 @@
 
 package org.apache.hc.core5.http.impl.bootstrap;
 
-import java.io.IOException;
```
ok2c: @clevertension Can we avoid moving imports around? It is not a big problem but it makes the change-set unnecessarily noisy.

httpcore5/src/main/java/org/apache/hc/core5/http/impl/bootstrap/AsyncRequester.java
```diff
@@ -132,4 +119,8 @@ public void close() throws IOException {
         ioReactor.close();
     }
 
+    @Override
```
ok2c: @clevertension Do we need to expose this method?

httpcore5/src/main/java/org/apache/hc/core5/reactor/DefaultConnectingIOReactor.java
```diff
@@ -119,28 +115,13 @@ public IOReactorStatus getStatus() {
     }
 
     @Override
-    public List<ExceptionEvent> getExceptionLog() {
-        return auditLog.isEmpty() ? Collections.<ExceptionEvent>emptyList() : new ArrayList<>(auditLog);
+    public IOWorkers.Selector getWorkerSelector() {
```
ok2c: @clevertension Why do we need to expose `IOWorkers.Selector`? Looks completely unnecessary to me.

httpcore5/src/main/java/org/apache/hc/core5/http/impl/bootstrap/AsyncRequester.java
clevertension: ok, i will adjust it

httpcore5/src/main/java/org/apache/hc/core5/http/impl/bootstrap/AsyncRequester.java
clevertension: this method must implement in the AsyncRequester, AsyncServer, because the 
AbstractAsyncServerAndRequester should use the IORector to connect, so change it to **protected** or **friendly**?

httpcore5/src/main/java/org/apache/hc/core5/reactor/DefaultConnectingIOReactor.java
clevertension: change it to **protected** or **friendly**?

ok2c:
@clevertension Committed with small changes as 2a14ba30fa155c75180b661976aeada583b6b816. Please review and close this PR.

ok2c:
@clevertension Please review and close.

PR #73, opened 2018-08-07
Fix HTTPCORE-549

clevertension:
```
mvn -Powasp -Dtest=false -DfailIfNoTests=false clean org.owasp:dependency-check-maven:3.3.1:aggregate
```
will generate report in **target/dependency-check-report.html**
image

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -350,8 +350,33 @@
           </reportSet>
         </reportSets>
       </plugin>
-
     </plugins>
   </reporting>
-
+  <profiles>
+    <profile>
+      <id>default</id>
+      <activation>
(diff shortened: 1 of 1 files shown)


ok2c:
@clevertension I think the parent project would be a better place for it. This way the plugin would need to be configured in each and every individual project. 

https://github.com/apache/httpcomponents-parent/

clevertension:
good suggestion, i will add and test it ASAP

clevertension:
done!!
https://github.com/apache/httpcomponents-parent/pull/1

clevertension:
ok, i have reviewed it, it is no problem, so close this PR

PR #74, opened 2018-08-21
HTTPCLIENT-1942: Add support for Reactive Streams

rschmitt:
This contribution adds initial support for the [Reactive Streams specification](http://www.reactive-streams.org/). The main part of this change is a pair of non-public classes, the `ReactiveDataProducer` and `ReactiveDataConsumer`, which are reactive adapters for Apache's AsyncDataProducer and AsyncDataConsumer interfaces. Two public classes are built on top of these types:

* `ReactiveEntityProducer`: An `AsyncEntityProducer` implementation backed by `ReactiveDataProducer` that allows a `Publisher<ByteBuffer>` to be streamed as a request body
* `ReactiveResponseConsumer`: An `AsyncResponseConsumer` that exposes the overall HTTP response as a `Message<HttpResponse, Publisher<ByteBuffer>>`, allowing the body to be consumed by a `Subscriber<ByteBuffer>`

These classes are enough to allow the async Apache client to straightforwardly integrate with reactive frameworks like [RxJava](https://github.com/ReactiveX/RxJava), [Akka](https://akka.io/), and [Vert.x](https://vertx.io/).

### DIFF
--- a/httpcore5-reactive/pom.xml
+++ b/httpcore5-reactive/pom.xml
@@ -1,0 +1,94 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ====================================================================
+  Licensed to the Apache Software Foundation (ASF) under one
+  or more contributor license agreements.  See the NOTICE file
+  distributed with this work for additional information
+  regarding copyright ownership.  The ASF licenses this file
+  to you under the Apache License, Version 2.0 (the
+  "License"); you may not use this file except in compliance
+  with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing,
+  software distributed under the License is distributed on an
+  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  KIND, either express or implied.  See the License for the
+  specific language governing permissions and limitations
+  under the License.
+  ====================================================================
+
+  This software consists of voluntary contributions made by many
+  individuals on behalf of the Apache Software Foundation.  For more
+  information on the Apache Software Foundation, please see
+  <http://www.apache.org />.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <artifactId>httpcore5-parent</artifactId>
+    <groupId>org.apache.httpcomponents.core5</groupId>
+    <version>5.0-beta4-SNAPSHOT</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>httpcore5-reactive</artifactId>
+  <name>Apache HttpComponents Core Reactive Extensions</name>
+  <description>Apache HttpComponents Reactive Streams Bindings</description>
+  <url>http://hc.apache.org/httpcomponents-core-ga</url>
+  <packaging>jar</packaging>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.apache.httpcomponents.core5</groupId>
+      <artifactId>httpcore5</artifactId>
+      <version>${project.version}</version>
+      <scope>compile</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.reactivestreams</groupId>
+      <artifactId>reactive-streams</artifactId>
+      <version>1.0.2</version>
+    </dependency>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>io.reactivex.rxjava2</groupId>
+      <artifactId>rxjava</artifactId>
+      <version>2.1.9</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.httpcomponents.core5</groupId>
+      <artifactId>httpcore5-h2</artifactId>
+      <version>5.0-beta3-SNAPSHOT</version>
+      <scope>compile</scope>
+    </dependency>
+  </dependencies>
+
+  <reporting>
+    <plugins>
+
+      <plugin>
+        <artifactId>maven-project-info-reports-plugin</artifactId>
+        <inherited>false</inherited>
+        <reportSets>
+          <reportSet>
+            <reports>
+              <report>dependencies</report>
+              <report>dependency-info</report>
+              <report>summary</report>
+            </reports>
+          </reportSet>
+        </reportSets>
+      </plugin>
+
+    </plugins>
+  </reporting>
+
+</project>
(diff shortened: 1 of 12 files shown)


httpcore5-reactive/pom.xml
```diff
@@ -0,0 +1,87 @@
...
+      <artifactId>reactive-streams</artifactId>
+      <version>1.0.2</version>
+    </dependency>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>io.reactivex.rxjava2</groupId>
```
ok2c: @rschmitt Can we remove compile dependency on `io.reactivex.rxjava2:rxjava` and make it test only or at least optional?

httpcore5-reactive/src/main/java/org/apache/hc/core5/reactive/ReactiveResponseConsumer.java
```diff
@@ -0,0 +1,176 @@
...
+    @Override
+    public void consumeResponse(
+        final HttpResponse response,
+        final EntityDetails entityDetails,
+        final HttpContext httpContext,
+        final FutureCallback<Void> resultCallback
+    ) {
+        this.entityDetails = entityDetails;
+        this.responseCompletion = new BasicFuture<>(resultCallback);
+        this.responseFuture.completed(new Message<HttpResponse, Publisher<ByteBuffer>>(response, reactiveDataConsumer));
```
ok2c: @rschmitt Why not using a simple `Consumer` instead? Do we really need another `Future` here?

httpcore5-reactive/src/main/java/org/apache/hc/core5/reactive/ReactiveDataProducer.java
```diff
@@ -0,0 +1,175 @@
...
+            publisher.subscribe(this);
+        }
+
+        final Throwable t = exception.get();
+        final Subscription s = subscription.get();
+        int buffersToReplenish = 0;
+        try {
+            synchronized (buffers) {
+                if (t != null) {
+                    // TODO: We need a reliable way to send RST_STREAM (without subsequent GOAWAY) on HTTP/2.
```
ok2c: @rschmitt By the way, why not using a subclass of `H2StreamResetException` here?

httpcore5-reactive/pom.xml
rschmitt: This was absolutely supposed to be test-only. I'll fix it.

httpcore5-reactive/src/main/java/org/apache/hc/core5/reactive/ReactiveDataProducer.java
rschmitt: That's an interesting option, but what happens if I throw it on HTTP/1.1?

ok2c: @rschmitt Nothing spectacular. I'll likely be handled like any other I/O error. I am still open to using a special annotation for exceptions, though, if you like that approach better.

httpcore5-reactive/src/main/java/org/apache/hc/core5/reactive/ReactiveResponseConsumer.java
rschmitt: I'm not sure what you mean by `Consumer`, because I thought this entire project was targeting JDK7. (RxJava provides a `Consumer<T>` interface, but the Reactive Streams API does not.) Either way, I would argue that we *do* need to return a `Future`, for the same reason that the `#execute` method returns a `Future`: callers need a single thing that they can block on (or decorate with callback logic) to wait for success, failure, or cancellation.

In the integration tests, there is a fairly consistent idiom:

```java
        final ReactiveResponseConsumer consumer = new ReactiveResponseConsumer();
        requester.execute(request, consumer, Timeout.ofSeconds(2), null);
        final Message<HttpResponse, Publisher<ByteBuffer>> response = consumer.getResponseFuture().get();
```

With a `Consumer`-based API, a caller might write:

```java
        final BlockingQueue<Message<HttpResponse, Publisher<ByteBuffer>>> streamingResponse =
            new ArrayBlockingQueue<>(1);
        final ReactiveResponseConsumer consumer =
            new ReactiveResponseConsumer(new Consumer<Message<HttpResponse, Publisher<ByteBuffer>>>() {
                @Override
                public void accept(final Message<HttpResponse, Publisher<ByteBuffer>> httpResponsePublisherMessage) {
                    streamingResponse.offer(httpResponsePublisherMessage);
                }
            });
        requester.execute(request, consumer, Timeout.ofSeconds(2), null);
        final Message<HttpResponse, Publisher<ByteBuffer>> response = streamingResponse.take();
```

However, the above code will block forever if an exception occurs and the response fails to come back. One way to fix this is by setting up a `CompletableFuture` or `BasicFuture` and then completing it from *both* callbacks:

```java
        final CompletableFuture<Message<HttpResponse, Publisher<ByteBuffer>>> streamingResponse =
            new CompletableFuture<>();
        final ReactiveResponseConsumer consumer =
            new ReactiveResponseConsumer(new Consumer<Message<HttpResponse, Publisher<ByteBuffer>>>() {
                @Override
                public void accept(final Message<HttpResponse, Publisher<ByteBuffer>> httpResponsePublisherMessage) {
                    streamingResponse.complete(httpResponsePublisherMessage);
                }
            });
        requester.execute(request, consumer, Timeout.ofSeconds(2), new FutureCallback<Void>() {
            @Override
            public void completed(Void result) {
                // TODO: Can this race with the other callback?
                streamingResponse.completeExceptionally(new RuntimeException("No stream returned"));
            }

            @Override
            public void failed(Exception ex) {
                streamingResponse.completeExceptionally(ex);
            }

            @Override
            public void cancelled() {
                streamingResponse.completeExceptionally(new CancellationException());
            }
        });
        final Message<HttpResponse, Publisher<ByteBuffer>> response = streamingResponse.join();
```

In short, constructing a correctly functioning `Future` for the streaming response is definitely something I want to do on the caller's behalf, rather than asking the caller to construct one manually from lower-level primitives.

httpcore5-reactive/src/main/java/org/apache/hc/core5/reactive/ReactiveDataProducer.java
rschmitt: I'll experiment with `H2StreamResetException`. The suggestion to use annotations is creative, but it may be overkill for what I need, which is probably not much.

rschmitt: `H2StreamResetException` works perfectly, except for the additional dependency edge from `:httpcore5-reactive` to `:httpcore5-h2`, which I think is acceptable.

ok2c: @rschmitt I'll refactor that bit once the change-set gets merged.

ok2c:
@rschmitt Committed to master as 1bae9643c912536793e820ac546730871e37155a. Please review and close this PR.

rschmitt:
Thanks for taking the time to review this contribution!

PR #79, opened 2018-09-08
Add HTTP status codes from RFC 5842

georgberky:
https://issues.apache.org/jira/browse/HTTPCORE-553


georgberky:
Test fails with OpenJDK 7 due to a timeout. Is this a problem we've caused?

PR #81, opened 2018-09-08
Add HTTP status codes from RFC 2774

georgberky:
https://issues.apache.org/jira/browse/HTTPCORE-556


georgberky:
Test fails with OpenJDK due to a timeout. Is this a problem we've caused?

georgberky:
Test fails with OpenJDK due to a timeout. Is this a problem we've caused?

ok2c:
Committed as 5da050c2b0d1b8c6c20c5098c0e4845e2b1278ac. Please review and close this PR.

PR #90, opened 2018-10-12
Define a timeout exception and deadline more precisely.

garydgregory:
Create and use a subclass of java.util.concurrent.TimeoutException called TimeoutValueException to precisely define a timeout exception.

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/concurrent/BasicFuture.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/concurrent/BasicFuture.java
@@ -95,7 +95,7 @@ public class BasicFuture<T> implements Future<T>, Cancellable {
         if (this.completed) {
             return getResult();
         } else if (waitTime <= 0) {
-            throw TimeoutValueException.ofMillis(msecs, msecs + Math.abs(waitTime));
+            throw TimeoutValueException.fromMillis(msecs, msecs + Math.abs(waitTime));
         } else {
             for (;;) {
                 wait(waitTime);
@@ -104,7 +104,7 @@ public class BasicFuture<T> implements Future<T>, Cancellable {
                 }
                 waitTime = msecs - (System.currentTimeMillis() - startTime);
                 if (waitTime <= 0) {
-                    throw TimeoutValueException.ofMillis(msecs, msecs + Math.abs(waitTime));
+                    throw TimeoutValueException.fromMillis(msecs, msecs + Math.abs(waitTime));
                 }
             }
         }
--- a/httpcore5/src/main/java/org/apache/hc/core5/concurrent/BasicFuture.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/concurrent/BasicFuture.java
@@ -33,6 +33,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 
 import org.apache.hc.core5.util.Args;
+import org.apache.hc.core5.util.TimeoutValueException;
 
 /**
  * Basic implementation of the {@link Future} interface. {@code BasicFuture}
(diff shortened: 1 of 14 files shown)


httpcore5/src/main/java/org/apache/hc/core5/concurrent/BasicFuture.java
```diff
@@ -94,7 +95,7 @@ public synchronized T get(final long timeout, final TimeUnit unit)
         if (this.completed) {
             return getResult();
         } else if (waitTime <= 0) {
-            throw new TimeoutException();
+            throw TimeoutValueException.ofMillis(msecs, msecs + Math.abs(waitTime));
```
ok2c: @garydgregory What is the value of this information? What good does it make to print out current time and deadline in milliseconds? I thought you were going to include the timeout value in the exception message, which would be of some marginal use.

garydgregory: @ok2c ,

Thank you for your comment.

What is the "timeout value"? Is it the boundary of time above which the timeout occurs or is it the actual elapsed time? To both avoid this confusion and to be as precise with the message as possible, both values are given: 

- The "deadline" in the message is the the boundary of time above which the timeout occurs.
- The "actual" in the message is the actual elapsed time.

Does that clarify things for you?

Gary

ok2c: @garydgregory To me it is Is it the _"boundary of time above which the timeout occurs"_. Yes, it does clarify but i personally do not find this information useful. The initial timeout value given by the user, say 30 seconds, might be marginally useful, but the number of milliseconds that have elapsed since midnight, January 1, 1970, when the operation timed out is not.

garydgregory: Hi @ok2c ,

The 'actual' value is _not_ the number of milliseconds that have elapsed since midnight, January 1, 1970. It is a value >= deadline relative to the deadline because 'msecs' is equal to the timeout value in millis. The 'actual' value is really 'msecs' plus how much over the deadline we waited represented by 'waitTime'. This is useful because if the deadline is 30 seconds, but we ended up waiting 5 minutes, then we or the user can look for what I claim would be a bug, depending on how the Future is used.

Gary

ok2c: @garydgregory This is clearly not how I interpret this piece of code.
```              
 final long now = System.currentTimeMillis();
 if (now > deadline) {
     leaseRequest.failed(TimeoutValueException.ofMillis(deadline, now));
```

ok2c:
@garydgregory Are going to re-spin this patch? Shall I delay the release because of it?

httpcore5/src/main/java/org/apache/hc/core5/concurrent/BasicFuture.java
garydgregory: Yes, for that call site, you are correct, my implementation is incorrect, I will adjust...

garydgregory: @ok2c I've updated this PR to fix the above issue and also refactored our use of a deadline long and its redundant code into a new Deadline class.

ok2c:
@garydgregory There is one test case failing in Travis CI with all JDKs. Could you please have a look?
I suggest this change gets included in the next release. I'll proceed with 5.0 BETA4 in the meantime.

garydgregory:
@ok2c With my latest commits, all Travis CI builds are green for this PR.

ok2c:
@garydgregory I'll merge this PR after HttpCore 5.0-beta4 gets tagged.

garydgregory:
@ok2c , OK, thank you.

ok2c:
Committed as b77420f. Please review and close this PR.

garydgregory:
PR merged; local build OK; closing.

PR #91, opened 2018-10-12
Interpret HTTP/2 window updates as increments

rschmitt:
The `CapacityChannel#update(int)` method accepts a capacity *increment*, essentially allowing the caller to signal that a certain number of bytes has just been freed up. However, the HTTP/2 implementation interprets that value as the *total* buffer capacity, and thus drops all capacity
increments below a certain size, causing the window size to irreversibly shrink over time.

This change causes HTTP/2 capacity updates to be applied directly to the input window as increments. Since there exist `AsyncEntityConsumer` implementations (such as `StringAsyncEntityConsumer`) that supply a capacity of `Integer.MAX_VALUE`, the `updateWindow` method has been modified to tolerate integer overflow, instead of throwing an ArithmeticException. This is intended as an interim measure until the topic of capacity channel update management can be addressed more comprehensively.

### DIFF
--- a/httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/AbstractHttp2StreamMultiplexer.java
+++ b/httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/AbstractHttp2StreamMultiplexer.java
@@ -193,7 +193,7 @@ abstract class AbstractHttp2StreamMultiplexer implements Identifiable, HttpConne
             HttpProcessor httpProcessor,
             BasicHttpConnectionMetrics connMetrics) throws IOException;
 
-    private int updateWindow(final AtomicInteger window, final int delta) throws ArithmeticException {
+    private int updateWindow(final AtomicInteger window, final int delta) {
         for (;;) {
             final int current = window.get();
             long newValue = (long) current + delta;
@@ -206,7 +206,7 @@ abstract class AbstractHttp2StreamMultiplexer implements Identifiable, HttpConne
             //TODO: needs to be removed
 
             if (Math.abs(newValue) > 0x7fffffffL) {
-                throw new ArithmeticException("Update causes flow control window to exceed " + Integer.MAX_VALUE);
+                newValue = Integer.MAX_VALUE;
             }
             if (window.compareAndSet(current, (int) newValue)) {
                 return (int) newValue;
@@ -215,7 +215,7 @@ abstract class AbstractHttp2StreamMultiplexer implements Identifiable, HttpConne
     }
 
     private int updateInputWindow(
-            final int streamId, final AtomicInteger window, final int delta) throws ArithmeticException {
+            final int streamId, final AtomicInteger window, final int delta) {
         final int newSize = updateWindow(window, delta);
         if (streamListener != null) {
             streamListener.onInputFlowControl(this, streamId, delta, newSize);
@@ -224,7 +224,7 @@ abstract class AbstractHttp2StreamMultiplexer implements Identifiable, HttpConne
     }
 
     private int updateOutputWindow(
-            final int streamId, final AtomicInteger window, final int delta) throws ArithmeticException {
+            final int streamId, final AtomicInteger window, final int delta) {
         final int newSize = updateWindow(window, delta);
         if (streamListener != null) {
             streamListener.onOutputFlowControl(this, streamId, delta, newSize);
@@ -368,16 +368,12 @@ abstract class AbstractHttp2StreamMultiplexer implements Identifiable, HttpConne
         return 0;
     }
 
-    private void updateInputCapacity(
-            final int streamId, final AtomicInteger inputWindow, final int inputCapacity) throws IOException {
-        if (inputCapacity > 0) {
-            final int streamWinSize = inputWindow.get();
-            final int chunk = inputCapacity - streamWinSize;
-            if (chunk > 0) {
-                final RawFrame windowUpdateFrame = frameFactory.createWindowUpdate(streamId, chunk);
-                commitFrame(windowUpdateFrame);
-                updateInputWindow(streamId, inputWindow, chunk);
-            }
+    private void incrementInputCapacity(
+        final int streamId, final AtomicInteger inputWindow, final int increment) throws IOException {
+        if (increment > 0) {
+            final RawFrame windowUpdateFrame = frameFactory.createWindowUpdate(streamId, increment);
+            commitFrame(windowUpdateFrame);
+            updateInputWindow(streamId, inputWindow, increment);
         }
     }
 
@@ -1365,7 +1361,7 @@ abstract class AbstractHttp2StreamMultiplexer implements Identifiable, HttpConne
             if (remoteEndStream) {
                 return;
             }
-            updateInputCapacity(id, inputWindow, increment);
+            incrementInputCapacity(id, inputWindow, increment);
         }
 
         @Override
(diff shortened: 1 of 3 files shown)


ok2c:
@rschmitt We cannot indiscriminately disregard update window overflow as that would be a direct violation of the HTTP/2 protocol. What we need to figure out is how to handle `CapacityChannel` increments gracefully in case they exceed the max limit.

ok2c:
@rschmitt How about this 277c628607e05c62d6fa9130a77ef7361aa6a49a?

rschmitt:
@ok2c That's an improvement, but I still don't think we've gotten at the root of the problem yet. I don't think that a hardcoded capacity increment of `Integer.MAX_VALUE` is an appropriate way of signaling unlimited capacity, but it's much more difficult under the current design to do the right thing. Shouldn't it be simple for a consumer to signal `Integer.MAX_VALUE` (or some more reasonable value) as the *initial* capacity, and thereafter to signal capacity increments at the end of every `consume` call? (Another subtlety: when you think about it, claiming that you have an infinite buffer is not the same thing as claiming to always process consumed data instantaneously.)

I think a few interface changes could simplify things quite a bit:

1. The `CapacityChannel` should be supplied exactly once, before the `consume` and `streamEnd` methods are ever called; this could be an `init` method, which also returns the initial capacity of the consumer.
2. The consumer should always signal capacity increments by calling `CapacityChannel`; the consumer's caller should never ask for an increment directly.

I'll see how difficult it is to implement these changes.

ok2c:
@rschmitt The current API has been designed to work well with both HTTP/1.1 and HTTP/2.0. For instance, the initial capacity is out of control for individual HTTP/2 consumers. They must be prepared to consume at least the initial amount of data defined by HTTP/2 settings during the HTTP/2 protocol handshake. The capacity value returned by `#consume` method was necessary to make the same API work with HTTP/1.1 connections that do not even support a concept of data capacity. There is no easy way for HTTP/1.1 consumers to make the event handler aware of the fact that they are not able to process more data. The only alternative I can think of using different capacity interfaces for different HTTP protocol versions which would be terrible and hardly any simpler.

Please do go ahead and try to simplify the APIs as you deem fit, but the result might well end up being a different API complex in a different way.

ok2c:
@rschmitt If the immediate problem has been resolved could you please close this PR and raise a new one once ready?

rschmitt:
@ok2c I've been meaning to circle back on this discussion. I did a few hours of work on this problem and got the majority of the integration tests passing with my alternate backpressure interface, but the changes are definitely invasive and destabilizing, simply because of how widely implemented these interfaces are within the codebase; I think that changing these interfaces requires either more time or more experience with the codebase than I currently have. However, since these are *user*-facing interfaces that we'll be stuck with after release, I think we should seriously consider biting the bullet and making the changes.

The interface I've been working with is:

```java
public interface AsyncDataConsumer extends ResourceHolder {
    int initCapacity(CapacityChannel capacityChannel);
    void consume(ByteBuffer src) throws IOException;
    void streamEnd(List<? extends Header> trailers) throws HttpException, IOException;
}
```

This interface implements the three points I suggested above, and is conceptually equivalent to the Reactive Streams API (where the `initCapacity` method is analogous to `onSubscribe`, and the `CapacityChannel` is the `Subscription`).

To address the points you made one by one:

> The current API has been designed to work well with both HTTP/1.1 and HTTP/2.0. For instance, the initial capacity is out of control for individual HTTP/2 consumers. They must be prepared to consume at least the initial amount of data defined by HTTP/2 settings during the HTTP/2 protocol handshake.

This is true, but I think it can be addressed simply by decoupling buffering and processing. Currently, if processing looks like:

`AbstractHttp2StreamMultiplexer -> Buffering processor`

We can simply refactor to:

`AbstractHttp2StreamMultiplexer -> Per-stream buffer -> AsyncDataConsumer`

This structure does a better job of separating internal concerns (per-stream buffering) and user concerns (the actual data and how it is processed), and makes it easier (fewer surprises) for users to implement the `AsyncDataConsumer` directly if they choose to do so.

> The capacity value returned by #consume method was necessary to make the same API work with HTTP/1.1 connections that do not even support the concept of data capacity.

I don't see how that follows. It's obviously true that HTTP/1.1 has no explicit windowing, but it's still an inherently asynchronous protocol, by virtue of the fact that it runs on top of TCP. You either read from the socket or abstain from reading from the socket, depending on your ability to process data (which is signaled via an asynchronous backpressure mechanism).

> There is no easy way for HTTP/1.1 consumers to make the event handler aware of the fact that they are not able to process any more input without data loss.

In reactive streaming, we signal exactly how much more data we *can* process; we don't explicitly tell the producer to *stop*, and the producer never sends us more data than we requested.

> The only alternative I can think of using different capacity interfaces for different HTTP protocol versions which would be terrible and hardly any simpler.

That would be terrible, but I think that protocol-specific buffering techniques (internal to the client) are all we really need here. The backpressure strategy I'm describing is one that I've implemented for both HTTP/1.1 and HTTP/2 on top of Netty; it works fine.

ok2c:
@rschmitt The proposed interface does look simpler and more aesthetically pleasing. If we can get it to work there will be no questions asked. 
> This is true, but I think it can be addressed simply by decoupling buffering and processing. 

True, but I suppose not every data stream may need buffering. This change would make it mandatory if I understand it correctly.

> In reactive streaming, we signal exactly how much more data we can process; we don't explicitly tell the producer to stop, and the producer never sends us more data than we requested.

This does sound reasonable and this is where we should probably start. 

When a particular change is too big and too disruptive what tends to work reasonably well is changing things one small thing at a time. Let's leave `AsyncDataConsumer` be for now but get rid of return type in AsyncDataConsumer#consume as the first step.

ok2c:
@rschmitt How about this for a start a5739b5dba9b10b357726602ede56406be2c8b69? The change-set still needs polish. I saw `Http1IntegrationTest#testLargePost` fail once, there is still might be a race condition somewhere but that could be a start.

rschmitt:
> True, but I suppose not every data stream may need buffering. This change would make it mandatory if I understand it correctly.

A buffering *layer* will be mandatory (it has to go somewhere in any HTTP/2 implementation), but data only has to be buffered if the downstream consumer is not ready to process it; otherwise it can be passed directly downstream. Either way, these details can be encapsulated by the client. (We can further reduce buffering by only buffering the *initial* window, and letting the downstream consumer drive capacity updates directly after it has been consumed.)

> Let's leave `AsyncDataConsumer` be for now but get rid of return type in AsyncDataConsumer#consume as the first step.

I was actually thinking the same thing; this change by itself is a significant simplification without having to tackle the more difficult "lifecycle" changes (where `updateCapacity` is redesigned along the lines of `onSubscribe`). a5739b5dba9b10b357726602ede56406be2c8b69 looks good to me; I say ship it!

ok2c:
> A buffering layer will be mandatory ..., but data only has to be buffered if the downstream consumer is not ready to process it;

@rschmitt Any notion of content buffering in the session layer makes me very, very uneasy. Once a chunk of message gets read from the socket channel, as far as the I/O selector is concerned the data has been fully consumed. It will no longer have any means of tracking that data. If the application layer stops reading data and clears interest in input events, the I/O selector will not be able to trigger read events on the associated channel, if the remaining message content has been buffered by the session. This problem has caused a lot of grief in HttpCore 4.x and its resolution involved a fair deal of ugly code. I was hoping to avoid having to do the same in HttpCore 5.x.

rschmitt:
@ok2c Your other option is to make minimum buffering requirements an explicit part of the contract, by doing one of the following:

1. Throwing an exception if an implementation of `initCapacity` returns an insufficient initial size
2. Passing your buffering requirements to the implementation as part of the `initCapacity` call and requiring the implementation to be able to buffer at least that much data

I don't have direct experience with the selector API in NIO, so it's tough for me to follow your description of the problem. Can't input event interest be set again as part of the `CapacityChannel#update` implementation (called back whenever data is read from the client's stream buffer)?

ok2c:
@rschmitt There are ways to approach the problem but I have not yet found an elegant solution. I need to step away from coding and force myself to write the documentation. At the moment the code is all yours. Do feel free to tweak the APIs as you see fit.

PR #98, opened 2018-11-30
SSLIOSession: Add `connectTimeout` constructor param

rschmitt:
This change adds low-level support for TLS handshake timeouts in the
class that actually performs the handshake. The contractual
`socketTimeout`, if set, will only be applied to the underlying
IOSession after the handshake is complete.


httpcore5/src/main/java/org/apache/hc/core5/reactor/InternalDataChannel.java
```diff
@@ -243,7 +243,8 @@ public void execute(final SSLIOSession sslSession) {
                         }
                     }
 
-                }))) {
+                },
+                null))) {
```
ok2c: @rschmitt Do we really need an extra constructor parameter if it is always set to `null`? Otherwise looks good.

rschmitt: The `null` argument here is temporary. Wiring this value up all the way from `InternalHttpAsyncExecRuntime` is going to require tons of additional changes.

PR #99, opened 2018-12-03
Add `handshakeTimeout` support throughout

rschmitt:


ok2c:
@rschmitt I was hoping to avoid having another type of timeout, but so be it. Android, for instance, also uses a distinct timeout value for TLS handshake. One kind request, though. Please also  `handshakeTimeout` to the classic `HttpRequester`. [This](https://github.com/apache/httpcomponents-core/blob/master/httpcore5/src/main/java/org/apache/hc/core5/http/impl/bootstrap/HttpRequester.java#L248) is likely to be the right place to apply the timeout.

rschmitt:
@ok2c To be clear, all I'm trying to do is use the existing `connectTimeout` as a TLS handshake timeout, without introducing more user-facing configuration in `RequestConfig`. As far as I can tell, these are the changes I need in order to propagate the `connectTimeout` from where it is (namely, `DefaultAsyncClientConnectionOperator#connect`) to where it needs to be (the `SSLIOSession` constructor).

garydgregory:
Hi @rschmitt ,
Are you worried about making ```RequestConfig``` more 'crowded' and confusing and are therefore piggybacking on top of another setting? Or, would the best design decision for users be to have a separate setting?
Gary

rschmitt:
@garydgregory I don't see it as piggybacking, I see it as the most natural option. An end-to-end connection (i.e. a route) can consist of many proxy hops, tunnels, layered/negotiated/upgraded protocols and so forth, and usually the route has to be established one hop at a time. It makes more sense to have a single `connectTimeout` that governs connection establishment (in the broadest sense).

The synchronous `CloseableHttpClient` does this already, at least with TLS handshakes. If nothing else, I'd like to bring the sync and async clients in line.

garydgregory:
That's all fine with me of course, but I'd want to make sure the Javadocs are updated so it is easy to understand what a timeout applies to but I did not check the state of Javadoc in this PR ;-)

PR #102, opened 2019-01-23
Merge connect and handshake timeouts in AbstractIOSessionPool

rschmitt:
This change essentially reverts a portion of 40cae2275 that appears to
have been unnecessary. Since the `requestTimeout` parameter is always a
connection timeout (at least in the client-side code), we don't need to
dedicate an additional parameter to the TLS handshake timeout value: we
simply use the `requestTimeout` (which has been renamed `connectTimeout`
for clarity).


rschmitt:
I'm not very confident about the rename from `requestTimeout` -> `connectTimeout`, because it appears that this parameter sometimes behaves as a `connectTimeout` and other times behaves as a `connectionRequestTimeout`. Suggestions and clarifications welcome!

ok2c:
@rschmitt `connectTimeout` sounds correct to me. I could not find any evidence of this variable being used as `connectionRequestTimeout` by `AbstractIOSessionPool`. Have I missed it somehow?

rschmitt:
I guess you're right. I suppose that makes sense, since this class is pooling sessions and not connections. (It's just a quirk of terminology that we also "connect" to establish a session.)

PR #105, opened 2019-02-13
[HTTPCORE-572] Move examples to the src/test folders for each module.

garydgregory:


pom.xml
```diff
@@ -185,7 +185,15 @@
               <outputDirectory>${basedir}/target/site/examples</outputDirectory>
               <resources>
                 <resource>
-                  <directory>src/examples</directory>
+                  <directory>src/test/java/org/apache/hc/core5/http/examples</directory>
```
ok2c: @garydgregory May I suggest that `maven-resources-plugin` section be moved to individual module poms? Module specific stuff in the main project pom looks a bit dodgy now.

garydgregory: @ok2c ,

Thank you for the review. I committed your requested changes.

Gary

ok2c:
@garydgregory Looks good. Could you please squash the commit though?

pom.xml
garydgregory: I must have messed up the squashing. I think it is easier to squash in GH when you merge. Can we do it that way?

PR #106, opened 2019-02-13
Immutable HttpEntity classes

ok2c:


httpcore5/src/main/java/org/apache/hc/core5/http/io/entity/InputStreamEntity.java
```diff
@@ -31,87 +31,52 @@
...
 import org.apache.hc.core5.util.Args;
 
 /**
- * A streamed, non-repeatable entity that obtains its content from
- * an {@link InputStream}.
+ * A streamed, non-repeatable entity that obtains its content from an {@link InputStream}.
  *
  * @since 4.0
  */
+@Contract(threading = ThreadingBehavior.IMMUTABLE_CONDITIONAL)
```
rschmitt: I don't really see how an `InputStream` wrapper could be regarded as immutable, unless `IMMUTABLE_CONDITIONAL` specifically means "immutable" in the Java Memory Model sense (all final fields, not made visible to other threads during construction), which has significantly weaker guarantees than true immutability.

httpcore5/src/main/java/org/apache/hc/core5/http/io/entity/SerializableEntity.java
```diff
@@ -59,29 +62,35 @@
...
-    public SerializableEntity(final Serializable ser, final boolean bufferize) throws IOException {
-        super();
+    public SerializableEntity(
+            final Serializable ser, final boolean bufferize, final ContentType contentType,
+            final String contentEncoding) throws IOException {
+        super(contentType, contentEncoding);
         Args.notNull(ser, "Source object");
         if (bufferize) {
             createBytes(ser);
+            this.objRef = null;
```
rschmitt: I don't understand what's happening here. Why are you nulling this field out after `createBytes` has assigned it?

httpcore5/src/main/java/org/apache/hc/core5/http/io/entity/SerializableEntity.java
```diff
@@ -91,33 +100,33 @@ private void createBytes(final Serializable ser) throws IOException {
     }
 
     @Override
-    public InputStream getContent() throws IOException, IllegalStateException {
+    public final InputStream getContent() throws IOException, IllegalStateException {
         if (this.objSer == null) {
             createBytes(this.objRef);
```
rschmitt: Based on this (apparent) lazy initialization path, this class is not currently immutable in any sense.

httpcore5/src/main/java/org/apache/hc/core5/http/io/entity/InputStreamEntity.java
ok2c: @rschmitt This is why there is `IMMUTABLE` and `IMMUTABLE_CONDITIONAL` with much weaker guarantees.

httpcore5/src/main/java/org/apache/hc/core5/http/io/entity/SerializableEntity.java
ok2c: @rschmitt Neither do it. I did not author this class.

httpcore5/src/main/java/org/apache/hc/core5/http/io/entity/SerializableEntity.java
ok2c: @rschmitt I removed `IMMUTABLE_CONDITIONAL` annotation from `ByteBufferEntity` and `InputStreamrEntity` and removed buffering code from `SerializableEntity`.

PR #109, opened 2019-02-16
Fixed broken generics in server bootstrap classes

ok2c:


ok2c:
@garydgregory Please let me know if there is anything else that needs fixing.

PR #110, opened 2019-02-19
HTTPCORE-568: Signal capacity in ReactiveDataConsumer whenever possible

rhernandez35:
This fixes a race condition in ReactiveDataConsumer that would cause the
input window to never be updated over HTTP/1.x connections. The bug
occurs when consume is the first method to call flushToSubscriber,
enough data has been received to exhaust the initial input window in the
stream duplexer, and there are still downstream requests for data that
haven't been served.

ReactiveDataConsumer attempts to flush data downstream whenever it gets
a chance, and only one thread is allowed to flush. After flushing, it
signals the number of bytes flushed to its capacity channel.

If the flush occurs because AbstractHttp1StreamDuplexer is calling
consume, it guarantees the capacity channel won't be set until consume
(and the flush) returns.  When there are still outstanding requests
(meaning the subscriber is unlikely to call request again), the flush
from consume will be the last attempt to signal capacity. The channel
isn't set yet, so nothing happens. As a result,
AbstractHttp1StreamDuplexer's input window remains empty,
ReactiveDataConsumer receives no further interactions, and the response
stops being read.

We can fix this by signaling capacity from both flushToSubscriber and
updateCapacity.

### DIFF
--- a/httpcore5-reactive/src/main/java/org/apache/hc/core5/reactive/ReactiveDataConsumer.java
+++ b/httpcore5-reactive/src/main/java/org/apache/hc/core5/reactive/ReactiveDataConsumer.java
@@ -54,10 +54,7 @@ import org.reactivestreams.Subscription;
 @Contract(threading = ThreadingBehavior.SAFE)
 final class ReactiveDataConsumer implements AsyncDataConsumer, Publisher<ByteBuffer> {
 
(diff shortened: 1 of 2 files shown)


rschmitt:
@ok2c PTAL

PR #111, opened 2019-02-22
Add a thread-safe capacity channel to AbstractHttp1StreamDuplexer

rhernandez35:
This solves two issues:

1. A data consumer that retains a reference to an old capacity channel
will no longer be able to modify state for future requests occurring on
the same connection (e.g. updating the window or requesting reads).

2. Coupling changes to the input window with changes to the io session
state prevents a race condition occuring when a consumer asynchronously
calls CapacityChannel#udpate at the same moment the input window is
exhausted by new input. Without synchronization, it was possible (albeit
unlikely) for the IO thread to decrease the window below zero, a
consumer thread to request input by calling CapacityChannel#update, then
the IO thread to clear the read request.

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/http/impl/nio/AbstractHttp1StreamDuplexer.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/http/impl/nio/AbstractHttp1StreamDuplexer.java
@@ -71,6 +71,7 @@ import org.apache.hc.core5.io.CloseMode;
 import org.apache.hc.core5.io.SocketTimeoutExceptionFactory;
 import org.apache.hc.core5.reactor.Command;
 import org.apache.hc.core5.reactor.EventMask;
+import org.apache.hc.core5.reactor.IOSession;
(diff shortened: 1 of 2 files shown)


rschmitt:
@ok2c PTAL

garydgregory:
How would you unit test this?

rhernandez35:
I can move CapacityWindow to its own class or make it package-private static and write some basic tests for it in isolation if you'd like. Testing _in situ_ would be much harder; I don't think there are any unit tests for any AbstractHttp1StreamDuplexer implementations.

garydgregory:
I think what matters here is the overall behavior. A black-box kind of test that fails without the main patch and passes with it.

ok2c:
@rhernandez35 @rschmitt Many thanks for this contribution. Could you please squash those commits into a single one? 

My biggest problem now is to understand whether the same changes need be made to HTTP/2 message multiplexer.

rhernandez35:
@ok2c It should be ok since each capacity channel is stream-scoped and ignores updates after receiving a frame with the END_STREAM flag set

PR #113, opened 2019-03-09
TLS improvements

ok2c:


httpcore5-testing/src/test/java/org/apache/hc/core5/testing/classic/ClassicServerAndRequesterTest.java
```diff
@@ -37,48 +39,66 @@
...
 
         @Override
         protected void before() throws Throwable {
             log.debug("Starting up test server");
             server = ServerBootstrap.bootstrap()
-                    .setSocketConfig(
-                            SocketConfig.custom()
-                                    .setSoTimeout(TIMEOUT)
-                                    .build())
+                    .setSslContext(scheme == URIScheme.HTTPS  ? SSLTestContexts.createServerSSLContext() : null)
```
garydgregory: Maybe refactor as `SSLTestContexts.createServerSSLContext(scheme)` to hide and reuse the ternary expression.

httpcore5/src/main/java/org/apache/hc/core5/http/nio/ssl/TlsSupport.java
```diff
@@ -0,0 +1,62 @@
...
+import org.apache.hc.core5.http.ssl.TlsCiphers;
+import org.apache.hc.core5.net.NamedEndpoint;
+import org.apache.hc.core5.reactor.ssl.SSLSessionInitializer;
+
+/**
+ * HTTP/1.1 TLS support methods
+ *
+ * @since 5.0
+ */
+public final class TlsSupport {
```
garydgregory: I've never been a fan of classes named *Util, *Support, and so on. It just feel like a potential kitchen sink. How about a name focused what support is actually implemented? I do not have a specific name in mind.

httpcore5/src/main/java/org/apache/hc/core5/http/ssl/TLS.java
```diff
@@ -0,0 +1,97 @@
...
+
+    public static ProtocolVersion parse(final String s) throws ParseException {
+        if (s == null) {
+            return null;
+        }
+        final ParserCursor cursor = new ParserCursor(0, s.length());
+        return TlsVersionParser.INSTANCE.parse(s, cursor, null);
+    }
+
+    public static String[] excludeWeak(final String[] protocols) {
```
garydgregory: Use `String...` IMO

httpcore5/src/main/java/org/apache/hc/core5/http/nio/ssl/TlsSupport.java
ok2c: @garydgregory I do not like them either but with no specific names in mind I am not sure what we can do here.

httpcore5/src/main/java/org/apache/hc/core5/http/ssl/TLS.java
ok2c: @garydgregory Done.

httpcore5/src/main/java/org/apache/hc/core5/http/nio/ssl/TlsSupport.java
garydgregory: Hi @ok2c ,

Since the class operates on our SSLSessionInitializer in a focused manner, one alternative would be to use a (Java 8, I know) default method in that interface. I know Java 8 is a separate discussion but I really do not see what we gain by keeping support for a dead stale version of Java like 7.

The alternative, since the method returns a new

ok2c: @garydgregory We already have plenty of `Support` and `Utils` classes. There is already `H2TlsSupport`. Treating `TlsSupport` differently would just make it worse in my option.

garydgregory: @ok2c OK, then all done.

httpcore5/src/main/java/org/apache/hc/core5/http/ssl/TLS.java
```diff
@@ -0,0 +1,97 @@
...
+            return null;
+        }
+        final List<String> enabledProtocols = new ArrayList<>();
+        for (final String protocol: protocols) {
+            if (!protocol.startsWith("SSL") && !protocol.equals(V_1_0.ident) && !protocol.equals(V_1_1.ident)) {
+                enabledProtocols.add(protocol);
+            }
+        }
+        if (enabledProtocols.isEmpty()) {
+            enabledProtocols.add(V_1_2.ident);
```
artem-smotrakov: Technically, there may be newer protocol versions available such as TLS 1.3 which may be considered as a better option. I am not sure if the method should figure out what is the latest protocol version available but maybe it should just return an empty array here. What do you think?

httpcore5/src/main/java/org/apache/hc/core5/http/ssl/TlsCiphers.java
```diff
@@ -0,0 +1,374 @@
...
+                    "TLS_RSA_WITH_AES_128_CCM_8",
+                    "TLS_RSA_WITH_AES_256_CCM_8",
+                    "TLS_PSK_WITH_AES_128_CCM",
+                    "TLS_PSK_WITH_AES_256_CCM",
+                    "TLS_PSK_WITH_AES_128_CCM_8",
+                    "TLS_PSK_WITH_AES_256_CCM_8"
+            )));
+
+    public static boolean isH2Blacklisted(final String cipherSuite) {
+        return H2_BLACKLISTED.contains(cipherSuite);
```
artem-smotrakov: Should the method use normalized strings?

httpcore5/src/main/java/org/apache/hc/core5/http/ssl/TLS.java
ok2c: @artem-smotrakov The trouble is that if  `enabledProtocols` is set to `null` that will prompt some JSSE provides to fall back onto the default set of protocols which may contain the same weak protocols we have just filtered out. We need to provide at least one value and presently `TLSv12` looks to be the most widely used strong protocol.

httpcore5/src/main/java/org/apache/hc/core5/http/ssl/TlsCiphers.java
ok2c: @artem-smotrakov Not sure how as those cipher names may contain mixed-case letters.

httpcore5/src/main/java/org/apache/hc/core5/http/ssl/TLS.java
artem-smotrakov: I see. TLSv12 seems to be a good choice. Higher protocol versions may be better but the method should be smarter to pick up a better option.

httpcore5/src/main/java/org/apache/hc/core5/http/ssl/TlsCiphers.java
artem-smotrakov: The JSSE docs and RFCs mention `_anon_` ciphers in lower case only but teoretically some TLS providers can use upper case. The method can expect it. That's also what you requested to change in https://github.com/apache/httpcomponents-client/pull/140#pullrequestreview-209877810

ok2c: @artem-smotrakov I remember that. I find apache/httpcomponents-client#140 to be a different story given that it employs pattern matching to filter out weak cipher. Here we have a list of specific ciphers.

PR #114, opened 2019-03-26
HTTPCORE-573 FileContentDecoder don't always enforce the maximum number of bytes to transfer

suiryc:
Limit the number of bytes read from the internal buffer when
transferring, so that the maximum requested ('count') is respected.

### DIFF
--- a/httpcore-nio/src/main/java/org/apache/http/impl/nio/codecs/IdentityDecoder.java
+++ b/httpcore-nio/src/main/java/org/apache/http/impl/nio/codecs/IdentityDecoder.java
@@ -93,8 +93,9 @@ public class IdentityDecoder extends AbstractContentDecoder
 
         long bytesRead;
         if (this.buffer.hasData()) {
+            final int maxLen = this.buffer.length();
             dst.position(position);
-            bytesRead = this.buffer.read(dst);
+            bytesRead = this.buffer.read(dst, count < maxLen ? (int)count : maxLen);
         } else {
             if (this.channel.isOpen()) {
                 if (position > dst.size()) {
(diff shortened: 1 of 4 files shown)


ok2c:
@suiryc Sounds reasonable. However would you also be willing to add unit tests covering the new functionality?

suiryc:
Sure I can at least give it a try.
I guess this is to be added in the TestLengthDelimitedDecoder and TestIdentityDecoder test classes.

suiryc:
For both implementations I added a test on limiting data transferred from the buffer and the source channel (the latter was already fine).
Squashed it in the original commit to keep the patch and the tests together.

httpcore-nio/src/main/java/org/apache/http/impl/nio/codecs/IdentityDecoder.java
```diff
@@ -93,8 +93,9 @@ public long transfer(
 
         long bytesRead;
         if (this.buffer.hasData()) {
+            final int maxLen = this.buffer.length();
             dst.position(position);
-            bytesRead = this.buffer.read(dst);
+            bytesRead = this.buffer.read(dst, count < maxLen ? (int)count : maxLen);
```
ok2c: @suiryc I think simple long to int cast here may not be good enough. I suppose it should rather be `(int) Math.min(count, Integer.MAX_VALUE)`

suiryc: I thought that casting was enough because it is only done when `count < maxLen` where `maxLen` is already an `int` (and thus should make `count <= Integer.MAX_VALUE`).
In this case do you think it's better to make it more explicit as you suggest ? (and also in `LengthDelimitedDecoder`)

ok2c: @suiryc You are absolutely right. Disregard my comment.

PR #115, opened 2019-04-02
HTTPCLIENT-1978: Filter characters before byte conversion

rschmitt:
This change causes ByteArrayBuffer to filter out all characters that
cannot be correctly converted to ISO-8859-1 by simple downcasting to a
byte. In addition, control characters and most whitespace characters are
filtered out. Filtered characters are replaced with a question mark,
except for surrogate pairs, which are replaced with two question marks.

https://en.wikipedia.org/wiki/ISO/IEC_8859-1
https://en.wikipedia.org/wiki/List_of_Unicode_characters


ok2c:
@rschmitt Looks good to me.

PR #117, opened 2019-04-04
Refactor APIs like failed(Exception) to failed(String, Exception)

garydgregory:
Better error reporting with this patch gives us:
```
Exception in thread "main" java.util.concurrent.ExecutionException:
Socket bind failure for socket ServerSocket[unbound],
address=0.0.0.0/0.0.0.0:8080, BacklogSize=0: java.net.BindException:
Address already in use: bind
	at org.apache.hc.core5.concurrent.BasicFuture.getResult(BasicFuture.java:74)
	at org.apache.hc.core5.concurrent.BasicFuture.get(BasicFuture.java:87)
	at org.apache.hc.core5.http.examples.AsyncFileServerExample.main(AsyncFileServerExample.java:165)
Caused by: java.net.BindException: Address already in use: bind
	at sun.nio.ch.Net.bind0(Native Method)
	at sun.nio.ch.Net.bind(Unknown Source)
	at sun.nio.ch.Net.bind(Unknown Source)
	at sun.nio.ch.ServerSocketChannelImpl.bind(Unknown Source)
	at sun.nio.ch.ServerSocketAdaptor.bind(Unknown Source)
	at org.apache.hc.core5.reactor.SingleCoreListeningIOReactor.processSessionRequests(SingleCoreListeningIOReactor.java:164)
	at org.apache.hc.core5.reactor.SingleCoreListeningIOReactor.processEvents(SingleCoreListeningIOReactor.java:102)
	at org.apache.hc.core5.reactor.SingleCoreListeningIOReactor.doExecute(SingleCoreListeningIOReactor.java:96)
	at org.apache.hc.core5.reactor.AbstractSingleCoreIOReactor.execute(AbstractSingleCoreIOReactor.java:81)
	at org.apache.hc.core5.reactor.IOReactorWorker.run(IOReactorWorker.java:44)
	at java.lang.Thread.run(Unknown Source)
Thu, 04 Apr 2019 20:17:06 GMT | HTTP server shutting down
```
Instead of just the BindException.

### DIFF
--- a/httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/AbstractHttp2IOEventHandler.java
+++ b/httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/AbstractHttp2IOEventHandler.java
@@ -54,7 +54,7 @@ class AbstractHttp2IOEventHandler implements HttpConnectionEventHandler {
         try {
             streamMultiplexer.onConnect(null);
         } catch (final HttpException ex) {
-            streamMultiplexer.onException(ex);
+            streamMultiplexer.onException(null, ex);
         }
     }
 
@@ -63,7 +63,7 @@ class AbstractHttp2IOEventHandler implements HttpConnectionEventHandler {
         try {
             streamMultiplexer.onInput();
         } catch (final HttpException ex) {
-            streamMultiplexer.onException(ex);
+            streamMultiplexer.onException(null, ex);
         }
     }
 
@@ -72,7 +72,7 @@ class AbstractHttp2IOEventHandler implements HttpConnectionEventHandler {
         try {
             streamMultiplexer.onOutput();
         } catch (final HttpException ex) {
-            streamMultiplexer.onException(ex);
+            streamMultiplexer.onException(null, ex);
         }
     }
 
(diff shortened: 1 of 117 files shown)


ok2c:
@garydgregory 117 changed files and a massive API change in order to have a slightly different error message? Is this really wise? Why not considering my suggestion of simply wrapping the original exception?

garydgregory:
Hi @ok2c,

```Why not considering my suggestion of simply wrapping the original exception?```

That's how I coded it initially but it feels like a hack and when you look at result it looks like a hack to me; YMMV.

The idea here IMO is that since we are still in beta, is to consider what is the _best_ API moving forward. So for me the breadth of the change is not important as long as the change is for the better.

The way I look at it is:
- When I call a failed() API and I have more to say than the exception, having to nest an exception feels weird when all I want is to give some information. 
- When I implement a `failed()` API and I get a message and an exception, I can format things maybe in a nicer way rather than just dumping the stack trace and assuming that message at the top is meaningful. If I do dump the stack trace, then I have yet another level of stack frames to look at, which is not so nice in this case.

I'm open to doing it either way, but my guess is that this is not the first time this will come up. In this specific case, it is something that's been bugging me forever, I just never took the time to do anything about it ;-)

Gary

garydgregory:
Hi @ok2c,

```Please consider changing #processSessionRequests only without breaking the existing APIs.```

Do you mean use a nested exception or only change the failed() method directly referenced from `processSessionRequests `.

Gary

ok2c:
@garydgregory I am afraid I will have to disagree on this being the best API. I think this change is redundant and its breadth shows how intrusive it is. 

There is no need to chain exceptions if you think it makes the stack trace harder to read. One can re-throw an exception with more context in it. Please see 335f59b97cbb098a45f3451d8c24c4504b377dc2

garydgregory:
@ok2c,

Like this then: https://github.com/apache/httpcomponents-core/pull/118

?

Gary

ok2c:
@garydgregory Yes, that would be my choice.

PR #119, opened 2019-04-12
Fix Timeout#ofMillis javadoc reference

rhernandez35:
Very small change, but it was bothering me.


garydgregory:
@rhernandez35,

Thank you for your PR! Keep them coming, we'd like to make 5.0 the best it can be :-)

Gary

PR #128, opened 2019-05-25
Added Greeting Example

rferreira:
Hello everyone, I'm not sure if there's any interest in this but here's a more sophisticated example of how to write a generic HTTP server using httpcore5.

Feel free to ignore it if you feel it's not needed.

### DIFF
--- a/httpcore5-h2/src/test/java/org/apache/hc/core5/http2/examples/Http2GreetingServer.java
+++ b/httpcore5-h2/src/test/java/org/apache/hc/core5/http2/examples/Http2GreetingServer.java
@@ -1,0 +1,195 @@
+/*
+ * ====================================================================
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ * ====================================================================
+ *
+ * This software consists of voluntary contributions made by many
+ * individuals on behalf of the Apache Software Foundation.  For more
+ * information on the Apache Software Foundation, please see
+ * <http://www.apache.org/>.
+ *
+ */
+package org.apache.hc.core5.http2.examples;
+
+import org.apache.hc.core5.function.Supplier;
+import org.apache.hc.core5.http.ContentType;
+import org.apache.hc.core5.http.EndpointDetails;
+import org.apache.hc.core5.http.EntityDetails;
+import org.apache.hc.core5.http.HttpException;
+import org.apache.hc.core5.http.HttpHeaders;
+import org.apache.hc.core5.http.HttpRequest;
+import org.apache.hc.core5.http.HttpResponse;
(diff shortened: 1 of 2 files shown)


httpcore5-h2/src/test/java/org/apache/hc/core5/http2/examples/Http2GreetingServer.java
```diff
@@ -0,0 +1,181 @@
...
+                            ContentType.APPLICATION_FORM_URLENCODED.getCharset());
+
+                    // composing greeting:
+                    greeting = String.format("Hello %s\n", args.get(0).getValue());
+                }
+
+                responseTrigger.submitResponse(new BasicResponseProducer(resp,
+                        new StringAsyncEntityProducer(greeting, ContentType.TEXT_PLAIN)), context);
+            } catch (Exception e) {
+                e.printStackTrace();
```
ok2c: @rferreira The exception handling code would also need to be reworked to something more presentable. Spitting exception stack traces to the console is not something we should be doing or recommending.

httpcore5-h2/src/test/java/org/apache/hc/core5/http2/examples/Http2GreetingServer.java
```diff
@@ -0,0 +1,181 @@
...
+                String greeting = "Hello stranger\n";
+
+                if (entityContentType.getMimeType().equals(ContentType.APPLICATION_FORM_URLENCODED.getMimeType())) {
+
+                    Charset charset = ContentType.APPLICATION_FORM_URLENCODED.getCharset();
+                    if (entityContentType.getCharset() != null) {
+                        charset = entityContentType.getCharset();
+                    }
+
+                    StringBuilder contents = new StringBuilder();
```
ok2c: @rferreira This way of parsing request entity content is not very efficient or even very inefficient. It is confusing to see the request entity content represented as a blocking `InputStream` only to be read line by line into a `StringBuilder` and then converted to `String`. I imagine you had your reasons for doing so but this may not necessarily be a very good generic example.

rferreira:
Yup, I'll revise this PR shortly.

httpcore5-h2/src/test/java/org/apache/hc/core5/http2/examples/Http2GreetingServer.java
rferreira: Yeah the rationale was that this was simpler for someone to grasp. I'll revise this with the String Entity Consumer and see how it looks.

rferreira:
@ok2c are these the correct checkstyle definition files? https://github.com/apache/httpcomponents-stylecheck

ok2c:
@rferreira Yes, they are. You should be looking at the version 2, though:
https://github.com/apache/httpcomponents-stylecheck/releases/tag/2

rferreira:
Ready for a second review @ok2c

httpcore5-h2/src/test/java/org/apache/hc/core5/http2/examples/Http2GreetingServer.java
```diff
@@ -0,0 +1,199 @@
...
+                            ContentType.APPLICATION_FORM_URLENCODED.getCharset());
+
+                    // composing greeting:
+                    greeting = String.format("Hello %s\n", args.get(0).getValue());
+                }
+
+                responseTrigger.submitResponse(new BasicResponseProducer(resp,
+                        new StringAsyncEntityProducer(greeting, ContentType.TEXT_PLAIN)), context);
+            } catch (final HttpException | UnsupportedCharsetException | IOException e) {
+                throw new IllegalStateException(e);
```
ok2c: @rferreira Much better, but I still dislike exception handling code. There is no need to re-throw  `HttpException` and  `IOException` as `IllegalStateException`. Those exceptions do not represent illegal service state. They are perfectly legal. Either handle them in your handler and return a valid HTTP message back to the caller, or let the framework do it for you.

rferreira:
All right, take 3. You know what they say, third time's the charm.

ok2c:
@rferreira Committed as 9dec68026a5607d75b90a53794db7e5f65ebd0c7 with some minor tweaks.

PR #131, opened 2019-06-14
Better HttpVersion and ProtocolVersion.

garydgregory:


httpcore5/src/main/java/org/apache/hc/core5/http/HttpVersion.java
```diff
@@ -81,36 +101,22 @@ public HttpVersion(final int major, final int minor) {
...
      *
      * @param major     the major version
      * @param minor     the minor version
      *
-     * @return  an instance of {@link HttpVersion} with the argument version
+     * @return an instance of {@link HttpVersion} with the argument version
+     * @throws IllegalArgumentException if either major or minor version number is negative
      */
     @Override
     public ProtocolVersion forVersion(final int major, final int minor) {
```
ok2c: @garydgregory I think `forVersion` method should be removed. It not being used anywhere in the productive code and has been rendered superfluous by static `get` method.

garydgregory: @ok2c I force pushed to remove the methods and another tweak.

PR #133, opened 2019-07-28
HTTPCORE-585 Avoid losing pending lease requests we can't fulfill

coxlinton:


ok2c:
@coxlinton Could you please add the JIRA ticket to the commit message and add some details as to what class or method are affected by the change? It will help me a great deal during the release process to have this information in the commit history.

coxlinton:
Done. Apologies for not doing so in the original commit.

PR #135, opened 2019-08-08
Execute Socket[Channel]#connect under doPrivileged

s1monw:
In order to allow users to run under a security manager that only grants
connect permission to the httpcore codebase the connect methods should
be executed in a doPriveledged block.
This is certainly not the only issue that users run into when they
install a SecurityManager with strict permissions but certainly the
most prominent and most likely one. Upstream components like the client
might also need to protect places accessing the proxy selector etc.

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/http/impl/bootstrap/HttpRequester.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/http/impl/bootstrap/HttpRequester.java
@@ -33,6 +33,9 @@ import java.io.OutputStream;
 import java.net.InetSocketAddress;
 import java.net.Proxy;
 import java.net.Socket;
+import java.security.AccessController;
+import java.security.PrivilegedActionException;
+import java.security.PrivilegedExceptionAction;
 import java.util.Set;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
@@ -82,6 +85,7 @@ import org.apache.hc.core5.pool.ManagedConnPool;
 import org.apache.hc.core5.pool.PoolEntry;
 import org.apache.hc.core5.pool.PoolStats;
 import org.apache.hc.core5.util.Args;
+import org.apache.hc.core5.util.Asserts;
 import org.apache.hc.core5.util.TimeValue;
 import org.apache.hc.core5.util.Timeout;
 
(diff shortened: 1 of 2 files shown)


s1monw:
for context, I ran into this issue while working on elasticsearch writing tests. We are a heavy user of this lib but also use a security manager. It requires some hacks on our end to use this library in such a context which is not the end of the world but would be nice to get fixed upstream, hence the PR, I am happy to work on any comments.

httpcore5/src/main/java/org/apache/hc/core5/http/impl/bootstrap/HttpRequester.java
```diff
@@ -261,7 +264,20 @@ private Socket createSocket(final HttpHost targetHost) throws IOException {
...
+            AccessController.doPrivileged(new PrivilegedExceptionAction<Object>() {
+                @Override
+                public Object run() throws IOException {
+                    sock.connect(targetAddress, socketConfig.getSoTimeout().toMillisIntBound());
+                    return null;
+                }
+            });
+        } catch (PrivilegedActionException e) {
+            // only checked exceptions are wrapped - error and RTExceptions are rethrown by doPrivileged
+            throw (IOException) e.getCause();
```
ok2c: @s1monw Even though `IOException` is the only checked exception thrown by the privileged operation, unconditional cast to  `IOException` still looks hideous. Could you please wrap it with `instance of` check and throw `IllegalStateException` in case the check fails or some such?

ok2c:
@s1monw Out of curiosity, are you using HttpCore or HttpClient and what version?

httpcore5/src/main/java/org/apache/hc/core5/http/impl/bootstrap/HttpRequester.java
s1monw: can we add an assertion instead? It seems like a broken behavior and should be fatal? WDYT?

s1monw:
> @s1monw Out of curiosity, are you using HttpCore or HttpClient and what version?

we are using httpclient v4.5.8, httpcore v4.4.11, httpasyncclient  v4.1.4 I understand that this change is on v5 and things have moved. Backporting is a different discussion I assume.

httpcore5/src/main/java/org/apache/hc/core5/http/impl/bootstrap/HttpRequester.java
ok2c: @s1monw Sure. That would serve. Please do use `org.apache.hc.core5.util.Asserts` for consistency with the rest of the code base.

s1monw: done

ok2c:
> we are using httpclient v4.5.8, httpcore v4.4.11, httpasyncclient v4.1.4 

@s1monw Curious. Were not folks at elasticsearch big time fans of Ning AsyncHttpClient at some point?

s1monw:
> @s1monw Curious. Were not folks at elasticsearch big time fans of Ning AsyncHttpClient at some point?

I think we evaluated this at some point in the past but we never used it. Since we have a http client we are using this lib and we intend to do so in the future.

ok2c:
@s1monw Could you please squash all changes into a single commit?

s1monw:
> @s1monw Could you please squash all changes into a single commit?

sure

s1monw:
@ok2c  done

s1monw:
@ok2c thanks for pulling this in. What's your strategy for backporting, is it worth it? I am happy to do it and open PRs where necessary. Lemme know

ok2c:
@s1monw Naturally I would like people to start migrating to 5.0, but I will merge a PR against 4.4.x if you submit one.

s1monw:
> @s1monw Naturally I would like people to start migrating to 5.0, but I will merge a PR against 4.4.x if you submit one.

 @ok2c I understand! The best I can do is to promise to move to 5 asap to support you folks. I still opened a backport of this in #138

ok2c:
@s1monw I suppose similar change (changes) should be made to `ConnectionSocketFactory` implementations in the HttpClient project in order to be consistent.

s1monw:
> @s1monw I suppose similar change (changes) should be made to ConnectionSocketFactory implementations in the HttpClient project in order to be consistent.

++ I opened a PR here https://github.com/apache/httpcomponents-client/pull/160

PR #137, opened 2019-08-12
HTTPCORE-589 Fix LaxConnPool to ensure pending requests assigned connections

coxlinton:
from available pool

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/pool/LaxConnPool.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/pool/LaxConnPool.java
@@ -37,6 +37,7 @@ import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicMarkableReference;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
 
 import org.apache.hc.core5.annotation.Contract;
 import org.apache.hc.core5.annotation.Experimental;
@@ -341,6 +342,8 @@ public class LaxConnPool<T, C extends ModalCloseable> implements ManagedConnPool
 
     static class PerRoutePool<T, C extends ModalCloseable> {
 
+        private enum RequestServiceStrategy { FIRST_SUCCESSFUL, ALL }
+
         private final T route;
         private final TimeValue timeToLive;
         private final PoolReusePolicy policy;
@@ -351,6 +354,7 @@ public class LaxConnPool<T, C extends ModalCloseable> implements ManagedConnPool
         private final Deque<LeaseRequest<T, C>> pending;
         private final AtomicBoolean terminated;
         private final AtomicInteger allocated;
+        private final AtomicLong releaseSeqNum;
 
         private volatile int max;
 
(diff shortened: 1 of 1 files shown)


httpcore5/src/main/java/org/apache/hc/core5/pool/LaxConnPool.java
```diff
@@ -392,17 +395,17 @@ public void shutdown(final CloseMode closeMode) {
...
             do {
                 prev = allocated.get();
                 next = (prev<poolmax)? prev+1 : prev;
             } while (!allocated.compareAndSet(prev, next));
-            return prev < next;
+            return (prev < next)? new PoolEntry<T,C>(route, timeToLive) : null;
         }
 
-        private void deallocatePoolEntry() {
+        private void disposePoolEntry(final PoolEntry<T, C> entry) {
```
ok2c: @coxlinton The `entry` parameter does not appear to be used and `disposePoolEntry` does not seem right given that no disposal actually takes place in this method. Can you think of a better name for this method?

httpcore5/src/main/java/org/apache/hc/core5/pool/LaxConnPool.java
```diff
@@ -481,12 +485,18 @@ public void release(final PoolEntry<T, C> releasedEntry, final boolean reusable)
...
+            releaseSeqNum.incrementAndGet();
             servicePendingRequest();
         }
 
+
         private void servicePendingRequest() {
+            servicePendingRequests(false);
+        }
+
+        private void servicePendingRequests(final boolean wholeQueue) {
```
ok2c: @coxlinton Let's use an enum here instead of a boolean.

garydgregory: - enum  QueueProcessingPolicy? QueueProcessingStrategy?
- PARTIAL vs. COMPLETE? 
- COMPLETE vs. INCOMPLETE?
- WHOLE vs. PARTIAL?

httpcore5/src/main/java/org/apache/hc/core5/pool/LaxConnPool.java
coxlinton: I intentionally created this method to be the matched pair to the concept of createPoolEntry: one method creates a new pool entry for the pool, the other is used for "the reverse of that"; the right description of that it the tricky thing :) Although the entry param currently isn't used, imagine a modification of the pool structure where we maintain a master list of all pool entries: then createPoolEntry would add the new entry to the master list, and disposePoolEntry would remove the entry from the master list. Would discardPoolEntry be any better?
But perhaps that's all unnecessary and I could just leave it as the previously declared deallocatePoolEntry - it's not really quite the exact analogue of createPoolEntry, but does accurately describe the current function of the method.

ok2c: @coxlinton I half suspected that as being the reason. I fully understand your desire to have a pair of symmetric allocation / deallocation methods. However, at present it just makes things more confusing. Those methods are not API visible and can be changed at point it becomes warranted. For now I would even go as far as removing `disposePoolEntry` and calling `allocated.decrementAndGet()` directly.

coxlinton: I've returned the method to it's former definition, and squashed the commits, so hopefully this is ready to merge now.

PR #138, opened 2019-08-14
Execute Socket[Channel]#connect under doPrivileged

s1monw:
In order to allow users to run under a security manager that only grants
connect permission to the httpcore codebase the connect methods should
be executed in a doPriveledged block.
This is certainly not the only issue that users run into when they
install a SecurityManager with strict permissions but certainly the
most prominent and most likely one. Upstream components like the client
might also need to protect places accessing the proxy selector etc.

Backport of #135

### DIFF
--- a/httpcore/src/main/java/org/apache/http/impl/pool/BasicConnFactory.java
+++ b/httpcore/src/main/java/org/apache/http/impl/pool/BasicConnFactory.java
@@ -183,13 +183,12 @@ public class BasicConnFactory implements ConnFactory<HttpHost, HttpClientConnect
         socket.setKeepAlive(this.sconfig.isSoKeepAlive());
         // Run this under a doPrivileged to support lib users that run under a SecurityManager this allows granting connect permissions
         // only to this library
-        final Socket finalSocket = socket;
         final InetSocketAddress address = new InetSocketAddress(hostname, port);
         try {
             AccessController.doPrivileged(new PrivilegedExceptionAction<Object>() {
                 @Override
                 public Object run() throws IOException {
-                    finalSocket.connect(address, BasicConnFactory.this.connectTimeout);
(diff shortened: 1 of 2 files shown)


ok2c:
@s1monw I tweaked `BasicConnFactory` a bit with 6f72e6c48e8e150d5771a70e7c405e34ce9a08e1. Could you please rebase your PR off the latest 4.4.x?

s1monw:
> Could you please rebase your PR off the latest 4.4.x? 

I can do that. I do wonder why you don't use githubs `squash and merge` from the merge button. I am not sure if it's the default for you but the little drop-down on the merge button gives you the option that way you don't get a merge commit and you don't need to rebase. you can also then edit the commit message?

ok2c:
@s1monw I could. But I would rather see and review the exact change-set to be committed and do `rebase and merge` instead. And I would also like to hope that you spot and eliminate `final Socket finalSocket = socket;` in the process of change-set review.

httpcore/src/main/java/org/apache/http/impl/pool/BasicConnFactory.java
```diff
@@ -177,7 +181,24 @@ public HttpClientConnection create(final HttpHost host) throws IOException {
             socket.setSoLinger(true, linger);
         }
         socket.setKeepAlive(this.sconfig.isSoKeepAlive());
-        socket.connect(new InetSocketAddress(hostname, port), this.connectTimeout);
+        // Run this under a doPrivileged to support lib users that run under a SecurityManager this allows granting connect permissions
+        // only to this library
+        final Socket finalSocket = socket;
```
ok2c: @s1monw Can we eliminate `finalSocket`? It is no longer required.

s1monw: done

PR #145, opened 2019-09-04
Added family property

behrangsa:
***NOTE***: The added code is taken almost verbatim from https://github.com/jax-rs/api/blob/master/jaxrs-api/src/main/java/javax/ws/rs/core/Response.java.

---

It is common in apps that use HTTP clients to check if a response status code is 2xx, 3xx, 4xx, 5xx, etc. and conditionally execute code based on that.

Spring's [HttpStatus](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/http/HttpStatus.html) for example, provides these methods:

```
public boolean is1xxInformational()
public boolean is2xxSuccessful()
public boolean is3xxRedirection()
public boolean is4xxClientError()
public boolean is5xxServerError()
```

as well as

```
public HttpStatus.Series series()
```

which returns an instance of [HttpStatus.Series](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/http/HttpStatus.Series.html).

If I am not mistaken, HTTP Components does not have a similar feature, hence the code in this PR *(which is is taken almost verbatim from https://github.com/jax-rs/api/blob/master/jaxrs-api/src/main/java/javax/ws/rs/core/Response.java)*.

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/http/message/StatusLine.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/http/message/StatusLine.java
@@ -46,13 +46,24 @@ public final class StatusLine implements Serializable {
 
     private static final long serialVersionUID = -2443303766890459269L;
 
-    /** The protocol version. */
+    /**
+     * The protocol version.
+     */
     private final ProtocolVersion protoVersion;
 
-    /** The status code. */
+    /**
+     * The status code.
+     */
     private final int statusCode;
 
-    /** The reason phrase. */
+    /**
+     * The status code class.
+     */
+    private final StatusClass statusClass;
+
+    /**
+     * The reason phrase.
+     */
     private final String reasonPhrase;
 
     public StatusLine(final HttpResponse response) {
@@ -60,26 +71,33 @@ public final class StatusLine implements Serializable {
         Args.notNull(response, "Response");
         this.protoVersion = response.getVersion() != null ? response.getVersion() : HttpVersion.HTTP_1_1;
         this.statusCode = response.getCode();
+        this.statusClass = StatusClass.from(this.statusCode);
         this.reasonPhrase = response.getReasonPhrase();
     }
+
     /**
      * Creates a new status line with the given version, status, and reason.
      *
-     * @param version           the protocol version of the response
-     * @param statusCode        the status code of the response
-     * @param reasonPhrase      the reason phrase to the status code, or
-     *                          {@code null}
+     * @param version      the protocol version of the response
+     * @param statusCode   the status code of the response
+     * @param reasonPhrase the reason phrase to the status code, or
+     *                     {@code null}
      */
     public StatusLine(final ProtocolVersion version, final int statusCode,
                       final String reasonPhrase) {
         super();
         this.statusCode = Args.notNegative(statusCode, "Status code");
+        this.statusClass = StatusClass.from(this.statusCode);
         this.protoVersion = version != null ? version : HttpVersion.HTTP_1_1;
         this.reasonPhrase = reasonPhrase;
     }
 
     public int getStatusCode() {
         return this.statusCode;
+    }
+
+    public StatusClass getStatusClass() {
+        return this.statusClass;
     }
 
     public ProtocolVersion getProtocolVersion() {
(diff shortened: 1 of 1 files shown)


garydgregory:
"which is is taken almost verbatim from https://github.com/jax-rs/api/blob/master/jaxrs-api/src/main/java/javax/ws/rs/core/Response.java"

I am not sure we can accept this PR due to Oracle's licensing terms. Someone will have to take the time to figure out the legal aspects. The simplest would be to completely redo the code as a clean room implementation.

e-hubert:
I'm not a licensing expert, but what about taking the code Oracle contributed and now maintained by the Eclipse foundation under Eclipse Public License v2: (https://github.com/eclipse-ee4j/jaxrs-api/blob/master/LICENSE.md) as a base instead?:
https://github.com/eclipse-ee4j/jaxrs-api/blob/master/jaxrs-api/src/main/java/jakarta/ws/rs/core/Response.java

ok2c:
@e-hubert We would still need to clarify whether the licensing terms are compatible with ASLv2. Someone will likely to have to write to legal@apache.org and ask for their guidance.

e-hubert:
Yeah, just wanted to add, that someone would still need to verify whether:
https://apache.org/legal/resolved.html#weak-copyleft-licenses 
"For small amounts of source that is directly consumed by the ASF product at runtime in source form, and for which that source is unmodified and unlikely to be changed anyway (say, by virtue of being specified by a standard), inclusion of appropriately labeled source is also permitted."
would apply here. 
Basically we are talking about an enum bound to RFC definitions, or?

ok2c:
@e-hubert @garydgregory  That would be good enough for me.

behrangsa:
The thing is there are only so many ways something this simple can be
reinvented. I will give it another try. :)

Best regards,
Behrang Saeedzadeh
blog.behrang.org

httpcore5/src/main/java/org/apache/hc/core5/http/message/StatusLine.java
```diff
@@ -89,6 +99,65 @@ public ProtocolVersion getProtocolVersion() {
...
+         */
+        OTHER;
+
+        /**
+         * Get the response status family for the status code.
+         *
+         * @param statusCode response status code to get the family for.
+         * @return family of the response status code.
+         */
+        public static Family familyOf(final int statusCode) {
```
michael-o: This should be `valueOf` of just `of`. The `family` is redundant.

behrangsa: > https://tools.ietf.org/html/rfc7231#section-6 talks about: the class of response
> 
> Maybe we should reuse this wording.

They are using `family` as `class` is overloaded in Java. We can use a synonym to family/class though (group, order, subclass, category, kind, variety, set, serie, etc.).

behrangsa: I will make the changes tomorrow morning -- it is 1:30 AM here. :)

garydgregory: Yes "class" is the proper name but I think we should call this enum `StatusClass` and put it in its own file.

michael-o: I agree with Gary here. Sound reasoning.

ok2c:
@behrangsa I would like to cut 5.0-beta9 this weekend. Will you be able to update the PR in time for BETA9?

behrangsa:
@garydgregory I made some changes to the PR per your comments.

httpcore5/src/main/java/org/apache/hc/core5/http/message/StatusLine.java
```diff
@@ -100,4 +118,84 @@ public String toString() {
...
+         */
+        SERVER_ERROR,
+
+        /**
+         * Any other HTTP status codes (e.g. non-standard status codes).
+         */
+        OTHER;
+
+        /**
+         * Get the response status class for the given status code.
```
garydgregory: "Get" -> "Gets"

httpcore5/src/main/java/org/apache/hc/core5/http/message/StatusLine.java
```diff
@@ -100,4 +118,84 @@ public String toString() {
...
+         *
+         * @param statusCode response status code to get the class for.
+         * @return class of the response status code.
+         */
+        public static StatusClass of(final int statusCode) {
```
garydgregory: I think "from" reads better than "of".

httpcore5/src/main/java/org/apache/hc/core5/http/message/StatusLine.java
behrangsa: Fixed. :)

httpcore5/src/main/java/org/apache/hc/core5/http/message/StatusLine.java
behrangsa: Fixed. :)

ok2c:
@behrangsa @garydgregory Please squash all commits into one either prior to `rebase and merge` or with `squash and merge`.

PR #146, opened 2019-09-05
Set accessiblity flag on reflected methods in ReflectionUtils

rhernandez35:
Prior to this change, attempts to reflectively invoke methods on
protected classes in other modules could fail on JDK9 and above.

Currently, the methods in ReflectionUtils are only used to get and set
properties related to ALPN on SSLEngine instances at runtime. On JDK11,
reflectively invoking the `getApplicationProtocol` method on
`SSLEngineImpl`, a protected class, throws an `IllegalAccessException`.
`callGetter` swallows this and returns `null`, causing the client
runtime to ignore whatever protocol was negotiated with ALPN and default
to HTTP/1.1.

This specific reflective access is still illegal and will eventually
break in a future Java release. A proper fix would require creating a
multi-release jar with a dedicated SSLEngine accessor class for JDK9+
that can natively invoke these methods.

### DIFF
--- a/httpcore5-testing/src/test/java/org/apache/hc/core5/testing/nio/H2AlpnTest.java
+++ b/httpcore5-testing/src/test/java/org/apache/hc/core5/testing/nio/H2AlpnTest.java
@@ -27,10 +27,6 @@
 
 package org.apache.hc.core5.testing.nio;
 
-import java.net.InetSocketAddress;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.Future;
-
 import org.apache.hc.core5.function.Supplier;
 import org.apache.hc.core5.http.ContentType;
 import org.apache.hc.core5.http.HttpException;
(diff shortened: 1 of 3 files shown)


ok2c:
@rhernandez35 Great stuff. Thank you. I will happily commit this change-set as is, however, is there a chance you could minimize stylistic changes to the existing classes? Is there is a particular problem with using JUnit rules? If so, can it be fixed consistently across the entire code base with a separate commit?

rhernandez35:
> is there a chance you could minimize stylistic changes to the existing classes?

Yeah, definitely. I hit everything with a round of vim auto-indent for speed after copy-pasting some stuff around, but I'll go back and do the formatting manually so it's consistent.

> Is there is a particular problem with using JUnit rules?

For test setup? No. the switch to just using `@Before` is an artifact from an earlier attempt at writing tests.

I'm not a fan of using rules for validating thrown exceptions. I think it's much safer to explicitly check for them where I expect them to be thrown in code. Otherwise, it's easy to accidentally skip assertions if a call unexpectedly throws the exception we're looking for.

ok2c:
@rhernandez35 Sure thing. All looks good. One last thing. All HttpClient classes use non-static `Logger` instances including test classes.  Can you keep it non-static for the consistency sake?

ok2c:
@rhernandez35 As far as the use of reflective calls goes, once 5.0 GA is out we can discuss whether or not we want to keep 5.1 branch Java 8 compatible or make Conscrypt TLS provider mandatory for Java 1.8 runtime. This would enable us to get rid of reflective calls altogether. 

Are you subscribed to the dev@hc.apache.org list?

rhernandez35:
@ok2 Yes, I am.

ok2c:
@rhernandez35 Good. I expect the issue of minimal JRE level to be raised at the dev list before the year end.

PR #148, opened 2019-09-10
HTTPCORE-599

ok2c:


ok2c:
@rhernandez35 @rschmitt @rhashimoto The fix itself is a trivial one liner: 84b001487e5216412f322ca90bada77c4a5aebb6 but I had to tweak `IOSession` API a bit to make it possible. Please take a look and let me know if you see anything disagreeable.

rschmitt:
@ok2c Looks okay to me, as long as this won't affect use cases like detecting closed connections and TCP RSTs.

ok2c:
@rschmitt Half-closed connections should not be problem but TCP RST indeed might. I am now very tempted to start a major rewrite of the buffering logic in the TLS/SSL layer

PR #151, opened 2019-09-24
TLS session and HTTP/1.1 data buffering improvements 

ok2c:


ok2c:
This is a big and potentially destabilizing change-set but it improves several critical aspects in the TLS session and HTTP/1.1 protocol handlers that always tended to be prone to rather nasty data buffering issues. Most commonly those issues manifest themselves as connections stuck with the last unprocessed chunk of data stuck in internal buffers but with no more input on the socket channel. I do hope that class of defects will be completely eliminated with this change-set.   

Change-log:
1. I/O event handlers can now proactively push data they have read ahead down to the event pipeline instead of expecting downstream handlers to pull that data from the session channel at some point.
1. No more application input data buffered in the SSL/TLS session layer
1. Less critical session locking in the SSL/TLS session layer
1. Simpler HTTP/1.1 input processing logic with just one processing loop
1. HTTP/1.1 event handler no longer needs to be "greedy" with input. 

@rschmitt @rhernandez35 Could you please see if this affects your code in any way? All tests pass for me without any modifications but please feel free to test this change-set internally.

@rhashimoto There is now a much better for HTTPCORE-599. Could you please see if the new code works for you as expected?

rschmitt:
@ok2c First, what happened to `AsyncRequestBuilder#setConfig`?

rschmitt:
@ok2c I wrote a class that implements both `AsyncRequestProducer` and `Configurable` so that I could supply config, but I'm still getting a ton of integration test failures which appear to be config-related.

ok2c:
@rschmitt Message builders got moved from client to core a release or two back. `RequestConfig` is a client specific construct, so I had to decouple it from the builders. One can still set `RequestConfig` for a particular message exchange by using `HttpClientContext#setConfig`.

rschmitt:
@ok2c Sorry, I still can't put two and two together with this API. How do I go from an `AsyncRequestBuilder`/`AsyncRequestProducer` to a `HttpClientContext`? How do I go from an `HttpRequest` (produced by `HttpClientContext#getRequest`) to an `AsyncRequestProducer` (consumed by various overloads of `CloseableHttpAsyncClient#execute`)?

ok2c:
@rschmitt That is one of doing it. One needs to pass `HttpContext` as a parameter to the `#execute` method.
```
final CountDownLatch latch1 = new CountDownLatch(1);
final SimpleHttpRequest request2 = SimpleHttpRequests.GET.create("http://httpbin.org/get");
final HttpClientContext clientContext = HttpClientContext.create();
clientContext.setRequestConfig(RequestConfig.custom()
        .setResponseTimeout(Timeout.ofSeconds(5))
        .build());
httpclient.execute(request2, clientContext, new FutureCallback<SimpleHttpResponse>() {

    @Override
    public void completed(final SimpleHttpResponse response2) {
        latch1.countDown();
        System.out.println(request2.getRequestUri() + "->" + response2.getCode());
    }

    @Override
    public void failed(final Exception ex) {
        latch1.countDown();
        System.out.println(request2.getRequestUri() + "->" + ex);
    }

    @Override
    public void cancelled() {
        latch1.countDown();
        System.out.println(request2.getRequestUri() + " cancelled");
    }

});
latch1.await();
```

rschmitt:
@ok2c Keep in mind that I'm running the latest commits of both core *and* client, with some local changes to get the latter to compile. With the latest `tls-nio` commit, I'm seeing two things of note:

* Connection timeout tests are failing because of an unexpected exception type being thrown (I believe this is an intentional change)
* Four tests involving trust store customization are failing (TCP reset from remote endpoint); further investigation is required

These results are actually pretty promising considering the extent of your changes. I'll continue investigating, but I think the `tls-nio` branch is basically correct.

ok2c:
> but I think the tls-nio branch is basically correct

@rschmitt I am really, really glad to hear that. I do think this change-set makes a number of things better and easier for us. I wish I had done those changes much earlier in the release process.

garydgregory:
That is what I like about our betas, we get the chance to get it right, all the while having useable artifacts in Maven Central 😀

rschmitt:
@ok2c I've figured out why those four TLS tests were failing: the `CLIENT_HELLO` message *only* lists AES-GCM cipher suites, and by some coincidence, the endpoint I'm calling doesn't support AES-GCM. This is only happening with the async client. Did the defaults change recently?

rschmitt:
Apparently, H2 cipher suite blacklisting is removing all the non-GCM cipher suites, in accordance with RFC 7540. The failing tests are fairly new, so this is presumably not a regression. Simply setting `HttpVersionPolicy.FORCE_HTTP_1` causes all the tests to pass.

@ok2c Congrats!

PR #152, opened 2019-09-27
validate connections everytime

hokuda:
I need a connetion pool which validates connections everytime when future.get() is called.
Invoking connpool.setValidateAfterInactivity(0) is an option to enable this.
Thanks in advance.

### DIFF
--- a/httpcore/src/main/java/org/apache/http/pool/AbstractConnPool.java
+++ b/httpcore/src/main/java/org/apache/http/pool/AbstractConnPool.java
@@ -251,7 +251,7 @@ public abstract class AbstractConnPool<T, C, E extends PoolEntry<T, C>>
                                 throw new ExecutionException(operationAborted());
                             }
                             final E leasedEntry = getPoolEntryBlocking(route, state, timeout, timeUnit, this);
-                            if (validateAfterInactivity > 0)  {
+                            if (validateAfterInactivity >= 0)  {
                                 if (leasedEntry.getUpdated() + validateAfterInactivity <= System.currentTimeMillis()) {
                                     if (!validate(leasedEntry)) {
                                         leasedEntry.close();


ok2c:
@hokuda This would change the default behavior of the class given that zero value is used to disable validation. What is the reason for not using a small positive value here?

hokuda:
@ok2c Thanks for looking into this! If a connection from the ConnPool is stale within validateAfterInactivity, an exception was raised. It seems it was introduced since 4.4, and I need the old one.ie, every connection must be validated. Should I add a field to ignore validateAfterInactivity rather than set it to 0? (how about "ignoreValidateAfterInactivity"?)

ok2c:
@hokuda I need to understand what it is exactly you are trying to accomplish before we make any API visible changes or break compatibility. 
Why can't you just use 1 ms instead of 0 ms? It is as _almost_ as likely that a connection can go stale immediately after passing the validation with 1 ms or 0 ms setting

hokuda:
@ok2c I'm 99.9999999 sure that 1ms works as expected but have to guarantee 100%.

ok2c:
@hokuda That is exactly the thing. 0 ms will **_not_** give you a 100% guarantee your code never ends up with a state connection as any connection can go stale _immediately_ after the successful validation but before the actual request execution.

hokuda:
@ok2c ah... you are right. Please close this.

LaunchPairs:
> @hokuda I need to understand what it is exactly you are trying to accomplish before we make any API visible changes or break compatibility.
> Why can't you just use 1 ms instead of 0 ms? It is as _almost_ as likely that a connection can go stale immediately after passing the validation with 1 ms or 0 ms setting

On our performance test environment, we still see a lot of cases with 1ms where the httpclient uses a stale TCP connection. If we use a version that always validates the connection, we haven't seen a single case yet. With this manually changed version, we completely ignore with the two checks:

`                    } else if (this.validateAfterInactivity > 0) { 
                        if (entry.getUpdated() + this.validateAfterInactivity <= System.currentTimeMillis())  {`

ok2c:
@LaunchPairs This might well be the case for your particular network constellation / application setup but the stale connection check is _inherently_ racy. There is absolutely no guarantee it can ensure 100% connection reliability in 100% of cases.

LaunchPairs:
We are aware of this, but I would still try to reduce the probability by eliminating this 1ms. We encountered this problem in connection with Nginx (Ingress) in a Kubernetes environment. Nginx very often breaks down keep-alive connections at each config reload. This produces avoidable unnecessary many errors on the apache HTTP client side.

ok2c:
@LaunchPairs From my experience there are better ways of dealing with the stale connections then the stack connection check, but I am fine with using a negative to disable connection validation instead of zero but preferably in 5.0 (master), not in 4.5.x.

LaunchPairs:
@hokuda I can make the pull request for the master but I think you want to resolve it in 4.x integration in WildFly / JBoss?

yvespp:
> @LaunchPairs From my experience there are better ways of dealing with the stale connections then the stack connection check, but I am fine with using a negative to disable connection validation instead of zero but preferably in 5.0 (master), not in 4.5.x.

Can you share your experiences dealing with stale connections?

We have the problem that we run into an exception because the connection was stale and the app doesn't know if it can be retried because it can't know if the request arrived at the backend or not. Also many services are not idempotent.

ok2c:
@yvespp The simplest and the most efficient way to have reliable connections is to limit their total time to live or to expire connections that have been idle beyond a given grace period.

ok2c:
@yvespp Pardon me. I should have said effective, not efficient. Some loss of efficiency can be expected due to not keeping alive connections that could potentially be re-used.

PR #153, opened 2019-10-22
Fixes HTTPCORE-609 by cancelling the key assigned to completed request

anuragagarwal561994:


garydgregory:
What if this.key is equal to the given key?

anuragagarwal561994:
@garydgregory I don't think it should create any issues, this is because key.cancel even on multiple calls doesn't produce an error. Check https://docs.oracle.com/javase/7/docs/api/java/nio/channels/SelectionKey.html#cancel()

The idea is that for a cancelled request if any key is assigned by any workflow it should be cancelled as well for maintaining consistent state.

PR #157, opened 2019-11-03
[HTTPASYNC-152] Adds state in SessionRequestImpl to avoid race condition in changing multiple variables

anuragagarwal561994:
Currently there are 2 states being maintained in SessionRequestImpl, terminated / completed.
Both of them are used differently, but one can be updated before the second is updated and read by the other's value. This may lead to race conditions which are prone to error.

One such example is issue https://issues.apache.org/jira/browse/HTTPASYNC-152.


ok2c:
@anuragagarwal561994 Good stuff. Thank you for contributing it.

anuragagarwal561994:
@ok2c when can we expect next release for 4.4.x branch?

ok2c:
@anuragagarwal561994 Likely this month.

PR #158, opened 2019-11-07
HTTPCORE-613: Adding staleConnectionCheckEnabled=true like semantic

pfrank13:
Making 0 a special value for validateAfterInactivity in which it will ALWAYS check if a connection is stale, much like the previous behavior when staleConnectionCheckEnabled was true


ok2c:
@pfrank13 We cannot change the current behavior in a patch release. It could only be changed in 4.5 (if we had one) or in 5.0. Please rebase your change-set off master.

ok2c:
Superseded by #160

PR #163, opened 2019-12-10
Allow for timeout while acquiring lock in StrictConnPool.

cwildman:
I mentioned in the mailing list that a connection request does not fail immediately when the time to acquire the StrictConnPool lock is longer than the requestTimeout. Instead it will only fail when the lock is finally acquired and the deadline is checked in processPendingRequest(). @ok2c agreed and suggested I propose a PR.

This PR changes StrictConnPool.lease to use the requestTimeout when attempting to grab the lock. If the lock can't be acquired within the timeout the request is failed with a DeadlineTimeoutException.

A few comments:

1. tryLock(timeout) is interruptible whereas lock() is not. I'm unclear if this would break any existing behavior.
2. I don't love that acquisition of the lock is captured in a boolean, it feels slightly dangerous if this block gets more complex down the road. However I chose this over nested try catches since we need to handle the InterruptedException.

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/pool/StrictConnPool.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/pool/StrictConnPool.java
@@ -173,11 +173,14 @@ public class StrictConnPool<T, C extends ModalCloseable> implements ManagedConnP
         Asserts.check(!this.isShutDown.get(), "Connection pool shut down");
         final Deadline deadline = Deadline.calculate(requestTimeout);
         final BasicFuture<PoolEntry<T, C>> future = new BasicFuture<>(callback);
-        boolean acquiredLock = false;
+        final boolean acquiredLock;
 
         try {
             acquiredLock = this.lock.tryLock(requestTimeout.getDuration(), requestTimeout.getTimeUnit());
-        } catch (final InterruptedException ignored) {
+        } catch (final InterruptedException interruptedException) {
+            Thread.currentThread().interrupt();
+            future.failed(interruptedException);
+            return future;
         }
 
         if (acquiredLock) {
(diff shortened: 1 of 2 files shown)


httpcore5/src/main/java/org/apache/hc/core5/pool/StrictConnPool.java
```diff
@@ -171,21 +171,33 @@ public void close() {
...
         try {
-            final LeaseRequest<T, C> request = new LeaseRequest<>(route, state, requestTimeout, future);
-            final boolean completed = processPendingRequest(request);
-            if (!request.isDone() && !completed) {
-                this.leasingRequests.add(request);
-            }
-            if (request.isDone()) {
-                this.completedRequests.add(request);
+            acquiredLock = this.lock.tryLock(requestTimeout.getDuration(), requestTimeout.getTimeUnit());
+        } catch (final InterruptedException ignored) {
```
ok2c: @cwildman `InterruptedException` handling logic does not look quite right to me. At the very least one must call                 `Thread.currentThread().interrupt()` when catching and not re-throwing `InterruptedException`. I also think we should either cancel request at that point or propagate the exception to the caller.

cwildman: Makes sense. I updated to set interrupted again and immediately return a failed future with the InterruptedException.

httpcore5/src/main/java/org/apache/hc/core5/pool/StrictConnPool.java
```diff
@@ -171,21 +171,36 @@ public void close() {
...
-            final boolean completed = processPendingRequest(request);
-            if (!request.isDone() && !completed) {
-                this.leasingRequests.add(request);
-            }
-            if (request.isDone()) {
-                this.completedRequests.add(request);
+            acquiredLock = this.lock.tryLock(requestTimeout.getDuration(), requestTimeout.getTimeUnit());
+        } catch (final InterruptedException interruptedException) {
+            Thread.currentThread().interrupt();
+            future.failed(interruptedException);
```
ok2c: @cwildman One last minor thing. Would not `future#cancel()` be more appropriate here?

cwildman: Yeah that does seem right.

PR #174, opened 2019-12-28
Development

ok2c:


ok2c:
@rschmitt Would you like to review the change-set adding support for `SETTINGS_MAX_HEADER_LIST_SIZE` H2 parameter?

httpcore5-h2/src/main/java/org/apache/hc/core5/http2/hpack/HPackDecoder.java
```diff
@@ -287,13 +290,25 @@ public Header decodeHeader(final ByteBuffer src) throws HPackException {
...
             final Header header = decodeHeader(src);
             if (header == null) {
                 break;
             }
+            if (enforceSizeLimit) {
+                listSize += header.getName().length() * 2;
+                if (header.getValue() != null) {
+                    listSize += header.getValue().length() * 2;
+                }
+                listSize += 32;
```
rschmitt: I don't like how this code duplicates logic and constants from `HPackHeader`. Also, why are you multiplying everything by `2`?

ok2c: @rschmitt How would you interpret the requirement below from section 6.5.2 about length in **octets**? As far as I understand `char` representation in Java takes two octets.

```The value is based on the
uncompressed size of header fields, including the length of the
name and value **in octets** plus an overhead of 32 octets for each
header field.
```

rschmitt: I am certain that the specification is referring to bytes on the wire, not the internal in-memory representation, which the peer cannot possibly know. Furthermore, due to [JEP 254](https://openjdk.java.net/jeps/254), modern JVMs now use one byte per character to represent most strings.

ok2c: @rschmitt Fair enough. Makes sense. Please review 9db8bf158441599de893d4f801dbb9ac71340eec

PR #191, opened 2020-01-28
Add org.apache.hc.core5.http.Method.normalizedValueOf(String) for use by

garydgregory:
httpclient5.

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/http/Method.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/http/Method.java
@@ -28,6 +28,7 @@
 package org.apache.hc.core5.http;
 
 import java.util.Locale;
+import java.util.Objects;
 
 /**
  * Common HTTP methods defined by the HTTP spec.
@@ -67,7 +68,7 @@ public enum Method {
             return false;
         }
         try {
-            return valueOf(value.toUpperCase(Locale.ROOT)).safe;
+            return normalizedValueOf(value).safe;
         } catch (final IllegalArgumentException ex) {
             return false;
         }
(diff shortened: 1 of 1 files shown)


httpcore5/src/main/java/org/apache/hc/core5/http/Method.java
```diff
@@ -78,12 +79,22 @@ public static boolean isIdempotent(final String value) {
...
     }
 
+    /**
+     * Returns the Method for a normalized {@code value} of a method name.
+     *
+     * @param method A method name like {@code "delete"}, {@code "DELETE"}, or any mixed-case variant.
+     * @return the Method for the given method name.
+     */
+    public static Method normalizedValueOf(final String method) {
+        return valueOf(Objects.requireNonNull(method, "method").toUpperCase(Locale.ROOT));
```
ok2c: @garydgregory Please use `Args.notNull` for consistency with the rest of the code base.

garydgregory: Done. But... oh why oh why are we _duplicating_ functionality that's in the JRE?

ok2c:
@garydgregory Looks good. Please squash the commits either prior to merging or when merging.

httpcore5/src/main/java/org/apache/hc/core5/http/Method.java
ok2c: > Done. But... oh why oh why are we _duplicating_ functionality that's in the JRE?

Because when Args had been introduced, there was no such functionality in JRE. And we have already discussed it. Many times. If you want to remove `Args` and `LangUtils` feel free to do so but do it _consistently_ across the _entire_ code base. Everywhere.

PR #195, opened 2020-04-17
Fix CharArrayBuffer.subSequence(beganIndex,endIndex) to return right result.

ufolr:
**CharArrayBuffer** implemented the **CharSequence** interface, 
and **CharSequence** define method:   
 `public CharSequence subSequence(final int beginIndex, final int endIndex)`
   
Threre are two problem in original implementation of `CharArrayBuffer.subSequence`

1. It call `CharBuffer.wrap(buffer.array, beginIndex, endIndex)` directly，but the 3rd parameter of the `CharBuffer.wrap` method is length, it means how many chars need to copy, not endIndex. It makes some problem such as pass CharArrayBuffer as a CharSequence in to Regex's matcher method like following:   
```CharArrayBuffer charSequence = new CharArrayBuffer(8);   
        charSequence.append("aabb123c");   
        Pattern p = Pattern.compile("\\D*(\\d+)\\D*");   
        Matcher m = p.matcher(charSequence);   
        Assert.assertEquals("123", m.find()?m.group(1):""); 
```

2. **CharArrayBuffer's** sub-sequence should return the result as same class of **CharArrayBuffer**. But' the old implementation use `CharBuffer.wrap` to make a result which instance of **HeapCharBuffer**. 

This problem exists in both 4.x and 5.x. I'd like see this problem to be fixed.

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/util/CharArrayBuffer.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/util/CharArrayBuffer.java
@@ -485,7 +485,7 @@ public final class CharArrayBuffer implements CharSequence, Serializable {
         if (beginIndex > endIndex) {
             throw new IndexOutOfBoundsException("beginIndex: " + beginIndex + " > endIndex: " + endIndex);
         }
-        return CharBuffer.wrap(this.array, beginIndex, endIndex);
+        return CharBuffer.wrap(this.array, beginIndex, endIndex - beginIndex);
     }
 
     @Override
--- a/httpcore5/src/test/java/org/apache/hc/core5/util/TestCharArrayBuffer.java
+++ b/httpcore5/src/test/java/org/apache/hc/core5/util/TestCharArrayBuffer.java
@@ -400,6 +400,16 @@ public class TestCharArrayBuffer {
     }
 
     @Test
+    public void testSubSequence() {
+        final CharArrayBuffer buffer = new CharArrayBuffer(16);
+        buffer.append(" name:  value    ");
+        Assert.assertEquals(5, buffer.indexOf(':'));
+        Assert.assertEquals(" name", buffer.subSequence(0, 5).toString());
+        Assert.assertEquals("  value    ",
+            buffer.subSequence(6, buffer.length()).toString());
+    }
+
+    @Test
     public void testSubSequenceIndexOfOutBound() {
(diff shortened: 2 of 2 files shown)


ok2c:
> CharArrayBuffer's sub-sequence should return the result as same class of CharArrayBuffer. But' the old implementation use CharBuffer.wrap to make a result which instance of HeapCharBuffer.

@ufolr Where is this requirement stated anywhere? I could not find any evidence of such requirement in `CharSequence` javadocs.

ufolr:
> > CharArrayBuffer's sub-sequence should return the result as same class of CharArrayBuffer. But' the old implementation use CharBuffer.wrap to make a result which instance of HeapCharBuffer.
> 
> @ufolr Where is this requirement stated anywhere? I could not find any evidence of such requirement in `CharSequence` javadocs. 


@ok2c There is no specification. Just my suggestion. I think its return should be intuitive. Such as sub noodle is a noodle lol.    
There maybe some method which accept a **CharArrayBuffer** as parameter, but can't take the sub-sequence result of **CharArrayBuffer**.
This will cause me confusion.    
Anyway, just my suggestion.

ok2c:
@ufolr While "sub noodle is a noodle" is conceptually nice, it does cause an unnecessary array allocation and an array copy, which is a very expensive operation. It completely defeats the purpose of having efficient access to a sub sequence of characters contained in the buffer.
Could we fix the bug without causing unnecessary buffer allocation?

ufolr:
@ok2c Yep. I saw that performance waste of performance. So I make **CharArrayBuffer**  can reuse the existing char array by adding an offset pos tag. Now, if there is no write option, the **CharArrayBuffer** will not do allocation or array copy. Just like an **HeapCharBuffer** do.   
Too do this, I add an constructor `CharArrayBuffer(final char[] b, final int off, final int len)` to `CharArrayBuffer`.

ok2c:
@ufolr I am sorry but this is too much so that "sub noodle could be a noodle". Let's just fix the bug and let the rest stay the same.

ufolr:
> @ufolr I am sorry but this is too much so that "sub noodle could be a noodle". Let's just fix the bug and let the rest stay the same.

@ok2c OK, I revert the changes about  "sub noodle", just fix the wrong result value problem.   
"sub noodle could be a noodle" is just a suggestion, and does not have to be modified.

ok2c:
@ufolr Cool! Many thanks for contributing the fix!

PR #196, opened 2020-04-20
HTTPCORE-627: URIBuilder::Adding query parameters result in removing …

netsrotr:
…the scheme specific part of an URI

Changed: URIBuilder to support new getter/setter to allow access to previously existing but private schemeSpecificPart variable. Idea was to open the class for more use cases, like RFC 6068. Issue discussion at [HTTPCORE-627](https://issues.apache.org/jira/browse/HTTPCORE-627)

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java
@@ -276,26 +276,26 @@ public class URIBuilder {
     }
 
     /**
-     * Sets the URI scheme specific part and append a list of parameters to this part.
+     * Sets the URI scheme specific part and append a variable arguments list of NameValuePair instance(s) to this part.
      *
      * @param schemeSpecificPart
-     * @param nvps Optional, can be null. List of NameValuePair query parameters to be reused by the specific scheme part
+     * @param nvps Optional, can be null. Variable arguments list of NameValuePair query parameters to be reused by the specific scheme part
      * @return this.
      * @since 5.1
      */
-    public URIBuilder setSchemeSpecificPartWithParameter(final String schemeSpecificPart, final NameValuePair... nvps) {
-        return setSchemeSpecificPartWithParameters(schemeSpecificPart, nvps != null ? Arrays.asList(nvps) : null);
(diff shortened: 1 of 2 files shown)


ok2c:
@netsrotr Looks good to me. Could you please add `@since 5.1` tags to the javadocs of the new methods and add `public URIBuilder setSchemeSpecificPart(final String schemeSpecificPart, final NameValuePair... nvps)` method as well?

httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java
```diff
@@ -264,13 +264,39 @@ public URIBuilder setScheme(final String scheme) {
...
+
+    /**
+     * Sets the URI scheme specific part and append a list of parameters to this part.
+     *
+     * @param schemeSpecificPart
+     * @param nvps Optional, can be null. List of NameValuePair query parameters to be reused by the specific scheme part
+     * @return this.
+     * @since 5.1
+     */
+    public URIBuilder setSchemeSpecificPartWithParameter(final String schemeSpecificPart, final NameValuePair... nvps) {
```
ok2c, line 286: @netsrotr `setSchemeSpecificPart` was a better name if you ask me. Why `setSchemeSpecificPartWithParameter`?

netsrotr: The overloads taking `NameValuePair...` and `List<NameValuePair>` cannot have the same name.
So I voted for an explicit setter (that match the getter) and differentiate the other by `...Parameter(...` and `...Parameters(...` 
But: I can take whatever you prefer... Suggestions?

ok2c: > The overloads taking NameValuePair... and List<NameValuePair> cannot have the same name
@netsrotr Are you sure? We already have `setPathSegments` overloaded methods.

netsrotr: Oops... ;-) you are right: the unit test points me to the wrong road... Will refine the PR. Sry

PR #204, opened 2020-07-21
HTTPCORE-638: SharedOutputBuffer must only call dataStreamChannel.endStream() once

malaysf:
It is possible for `flush` to be called after `markCompleted` and both would call `dataStreamChannel.endStream` which is incorrect, `endStream` must only be called once.


httpcore5/src/main/java/org/apache/hc/core5/http/nio/support/classic/SharedOutputBuffer.java
```diff
@@ -162,4 +164,11 @@ private void waitFlush() throws InterruptedIOException {
         setInputMode();
     }
 
+    private void propagateEndStreamOnce() throws IOException {
```
ok2c, line 167: @malaysf One really minor thing. I suppose just `propagateEndStream` as the method name should be clear enough.

malaysf: Sure, I'll make the change shortly

PR #214, opened 2020-08-06
Fix MonitoringResponseOutOfOrderStrategyIntegrationTest on new JREs

carterkozak:
The timing changed around jre13, it seems that plain http
detection using InputStream.available does not detect responses
as quickly as the older implementation prior to jep353.
The test has been updated to use smaller write buffers and
force a flush after each write.


ok2c:
Folded into 8838ad028

PR #216, opened 2020-08-12
HTTPCORE-645: Increase blocking default chunk size to 8 KiB and reuse chunk buffers

carterkozak:
This PR includes two commits, because this fix is implemented in two parts:
1. Reuse chunk buffers between requests using a persistent connection in order to eliminate the performance cost of allocating buffers
2. Increase the default value from 2 KiB to 8 KiB matching the default buffer size.

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/http/impl/io/BHttpConnectionBase.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/http/impl/io/BHttpConnectionBase.java
@@ -158,7 +158,7 @@ class BHttpConnectionBase implements BHttpConnection {
     private byte[] getChunkedRequestBuffer() {
         if (chunkedRequestBuffer == null) {
             final int chunkSizeHint = this.http1Config.getChunkSizeHint();
-            chunkedRequestBuffer = new byte[chunkSizeHint > 0 ? chunkSizeHint : 2048];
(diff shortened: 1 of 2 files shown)


httpcore5/src/main/java/org/apache/hc/core5/http/impl/io/BHttpConnectionBase.java
```diff
@@ -147,13 +149,20 @@ protected OutputStream createContentOutputStream(
...
+            return new ChunkedOutputStream(buffer, outputStream, getChunkedRequestBuffer(), trailers);
         } else {
             return new IdentityOutputStream(buffer, outputStream);
         }
     }
 
+    private byte[] getChunkedRequestBuffer() {
+        if (chunkedRequestBuffer == null) {
+            final int chunkSizeHint = this.http1Config.getChunkSizeHint();
+            chunkedRequestBuffer = new byte[chunkSizeHint > 0 ? chunkSizeHint : 8192];
```
carterkozak, line 161: I think this should still be `>= 0` to avoid changing behavior when zero is set, allowing each write to produce a chunk.

carterkozak: On second review this matches the logic in ChunkedOutputStream which used the default buffer size when chunkSizeHint was set to zero.

PR #221, opened 2020-08-27
HTTPCORE-649: Implement ByteArrayBuffer.append(ByteBuffer)

carterkozak:


httpcore5/src/main/java/org/apache/hc/core5/http/nio/entity/BasicAsyncEntityConsumer.java
```diff
@@ -60,16 +60,7 @@ protected int capacityIncrement() {
...
-            return;
-        }
-        if (src.hasArray()) {
-            buffer.append(src.array(), src.arrayOffset() + src.position(), src.remaining());
-        } else {
-            while (src.hasRemaining()) {
-                buffer.append(src.get());
-            }
-        }
+        buffer.append(src);
```
carterkozak, line 63: This resolves an inconsistency where previously the src buffer position would advance only if the buffer was not array backed.

ok2c (approved):

PR #223, opened 2020-09-14
Update ckozak email address in 5.1 release notes

carterkozak:


ok2c:
@carterkozak C'mon, just push it to the branch, would you? No need for open a PR for stuff like that.
By the way, would you like to change your email address in the commits as well?

carterkozak:
Sure, I'll do both.

PR #224, opened 2020-09-25
Prevent an output request in AbstractHttp1StreamDuplexer from being erased when it occurs simultaneously with a call to onOutput

cweld510:
I’ve observed a deadlock when writing to a ContentOutputStream obtained through AbstractClassicEntityProducer. We’ve looked through the layers at how ContentOutputStream’s SharedOutputBuffer coordinates between threads and traced this to a race condition in AbstractHttp1StreamDuplexer. The issue is that AbstractHttp1StreamDuplexer.requestSessionOutput can be called, but the event loop never calls onOutput, which then doesn’t call produceOutput.

When a thread calls requestSessionOutput in the duplexer, it increments an AtomicInteger called outputRequests (used to track the current number of requests for output), and sets an OP_WRITE event. An event thread notices the OP_WRITE state, and responds to it by calling AbstractHttp1StreamDuplexer.onOutput(). This method then calls AbstractHttp1StreamDuplexer.produceOutput(). After the event thread returns, if requestSessionOutput has not been called since the event thread returned from produceOutput, the thread clears the event and sets outputRequests to zero. Otherwise, it decrements outputRequests after it finishes.

How this is going wrong: Let’s say that AbstractHttp1StreamDuplexer.outputRequests starts with a value of 1 and that the event thread is in the middle of a call to onOutput and has just returned from calling produceOutput. At that moment, a call to requestSessionOutput increments the outputRequests counter to 2. Then line 350: “final int pendingOutputRequests = outputRequests.get();” pendingOutputRequests is set to 2, and then we set pendingOutputRequests back to zero and unset the OP_WRITE state in lines 356 and 357:
                if (!outputPending && !outbuf.hasData() && outputRequests.compareAndSet(pendingOutputRequests, 0)) {
                    ioSession.clearEvent(SelectionKey.OP_WRITE);

Because the OP_WRITE event was cleared, onOutput (and thus produceOutput) will not be called again, even though a caller requested one.

The simplest way to fix this problem is to set pendingOutputRequests before the call to produceOutput; this makes the condition on line 356 false whenever requestSessionOutput is called directly after produceOutput. This change is included in my pull request.

I wasn’t able to reproduce the issue in a unit test because AbstractHttp1StreamDuplexer has several dependencies (e.g. the behavior of event threads) that I couldn’t figure out how to mock well, and the class doesn’t have any existing unit tests. A test in my company’s codebase that reliably reproduced the race condition passes after the change, but failed before, and I have verified that the change works as expected by using my debugger. However, I’d be happy to create a unit test to verify the correctness of this change, if there is a way to do so effectively.

Does this seem like the correct fix?

Thank you! --Colin

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/http/impl/nio/AbstractHttp1StreamDuplexer.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/http/impl/nio/AbstractHttp1StreamDuplexer.java
@@ -346,8 +346,8 @@ abstract class AbstractHttp1StreamDuplexer<IncomingMessage extends HttpMessage, 
             ioSession.getLock().unlock();
         }
         if (connState.compareTo(ConnectionState.SHUTDOWN) < 0) {
-            produceOutput();
             final int pendingOutputRequests = outputRequests.get();
+            produceOutput();
             final boolean outputPending = isOutputReady();
             final boolean outputEnd;
             ioSession.getLock().lock();


ok2c:
@cweld510 I will look into this issue. Could you please post a thread dump exhibiting the deadlock, though, for my reference?

cweld510:
Here's a thread dump I got from pausing IntelliJ during the deadlock:
[threads_report.txt](https://github.com/apache/httpcomponents-core/files/5285321/threads_report.txt)

For context, the first thread (bzWebTest@12788) is the thread running our test. We execute a call using a variant of AbstractClassicEntityProducer that exposes a ContentOutputStream. Our test thread writes to the output stream. I've removed the part of the callstack that relates to our code.

The output stream wraps an output buffer of type BzSharedOutputBuffer. This class is completely identical to SharedOutputBuffer; we made it to incorporate [this fix](https://github.com/apache/httpcomponents-core/pull/204) because it hadn't been released yet.

ok2c:
@cweld510 
1. It is not a thread deadlock in its true sense. Apparently some connections stop sending outgoing data due to a race condition.
2. What bothers me is why an entity producer would signal output readiness and have no output data at the same time. This does not sound quite right to me. Can you please find out why `final boolean outputPending = isOutputReady()` returns false?

cweld510:
@ok2c 

You're right that calling this a deadlock is a misnomer. I can change my description.

In response to your second question: when the race condition occurs, the thread that calls requestOutput from the SharedOutputBuffer doesn't write to the output buffer owned by that class, e.g. by calling `buffer().put(src)`. Instead, it calls `dataStreamChannel.write(src)` on line 98 of SharedOutputBuffer, but writes zero bytes. So `final boolean outputPending = isOutputReady()` will be false in this circumstance, because `isOutputReady()` only checks the buffer owned by SharedOutputBuffer.

To provide more information, I'm attaching a callstack of the thread that calls `dataStreamChannel.write(src)` on line 98 of SharedOutputBuffer. (line numbers are slightly different in BzSharedOutputBuffer but the `write` method is identical.) Hopefully this can show how a thread can signal output readiness but `final boolean outputPending = isOutputReady()` and `outbuf.hasData()` will both be false. The callstack isn't entirely complete, because i just stepped into `dataStreamChannel.write(src)` several layers deep and eventually copied it, but it should still outline the write thread's path of execution. Note that `SSLIOSession.write()` can return 0 here.


[stacktrace.txt](https://github.com/apache/httpcomponents-core/files/5287463/stacktrace.txt)

ok2c:
@cweld510 What is very important for me to establish whether it is conceptually legal for a data producer to signal output interest and at the same time report zero available data

cweld510:
@ok2c 
When the race condition occurs and the data producer signals output interest, there is some data in the `outEncrypted` buffer in `SSLIOSession`. It's the presence of this data that causes `dataStreamChannel.write(src)` to return 0 in the circumstance I described in my previous comment. After the IO Reactor thread calls `AbstractHttp1StreamDuplexer.onOutput`, it will flush the data from the SSLIOSession's buffer to the socket. So there is some available data at the time the data producer signals output readiness. It might not be what you are thinking about, though.

ok2c:
@cweld510 Merged to master and 5.0.x.

PR #226, opened 2020-10-16
Add workaround of java nio bug 

horizonzy:
Recently, I found my web programer cpu spike. finnly I found it should be java nio bug. refer eclipse/jetty.project#2205. I notice in my project, the logic in `org.apache.http.impl.nio.reactor.AbstractIOReactor#execute`,  exist `                        readyCount = this.selector.select(selectTimeout);`. This is the reason of cpu spike. So I want add a way to workaround it.
 I am not good at nio code, this is my clumsy code, I hope it will solve the cpu spike problem. If it can't work normally, can community create a pr to fix it? Thanks very much.

### DIFF
--- a/httpcore-nio/src/main/java/org/apache/http/impl/nio/reactor/AbstractIOReactor.java
+++ b/httpcore-nio/src/main/java/org/apache/http/impl/nio/reactor/AbstractIOReactor.java
@@ -35,10 +35,7 @@ import java.nio.channels.ClosedSelectorException;
 import java.nio.channels.SelectionKey;
 import java.nio.channels.Selector;
 import java.nio.channels.SocketChannel;
-import java.util.Collections;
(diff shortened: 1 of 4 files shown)


ok2c:
@horizonzy Not only this hack is incredibly hideous, there is no convincing evidence that it actually solves the problem. Is there a reproducer for this issue?

horizonzy:
> @horizo​​nzy不仅这种骇人听闻的骇人听闻，而且还没有令人信服的证据表明它确实可以解决问题。是否有针对此问题的复制器？

Hi,ok2c 😄. Thanks for you notice this pr. I report it in jara(https://issues.apache.org/jira/projects/HTTPCORE/issues/HTTPCORE-650?filter=allopenissues). I apply some information in it. If you need more info, just refer me.

PR #230, opened 2020-11-22
Bug fix: Fixed ignoring maxResultLength in toString method of EntityUtils

Klawru:
Currently, the EntityUtils.toString method ignores the maxRequestLength parameter.


ok2c:
@Klawru Ported to master as well.

PR #234, opened 2020-12-04
Add appendPath and appendPathSegments methods to URIBuilder

larshelge:
This PR is adding `appendPath` and `appendPathSegments` methods to `URIBuilder`.

The added benefit of these methods is allowing a URI to be built successively throughout an application. As an example, one method could set the base URL, pass the `URIBuilder` along to the next method which can set an API version, and have a third method set the context path.

This functionality does not fit nicely in an external utility method, as it breaks the builder pattern/fluent style.

This would be equivalent to Spring's [UriComponentsBuilder # path](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/util/UriComponentsBuilder.html#path-java.lang.String-). When migrating a recent application from Spring to HTTP Components, an equivalent for this method was lacking.

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java
@@ -505,9 +505,7 @@ public class URIBuilder {
      * @return this.
      */
     public URIBuilder appendPath(final String path) {
-        final List<String> segments = new ArrayList<>(getPathSegments());
-        segments.addAll(path != null ? splitPath(path) : new ArrayList<String>());
-        setPathSegments(segments);
+        appendPathSegments(path != null ? splitPath(path) : new ArrayList<String>());
         return this;
     }
 
@@ -530,10 +528,7 @@ public class URIBuilder {
      * @return this.
      */
     public URIBuilder appendPathSegments(final String... pathSegments) {
-        final List<String> segments = new ArrayList<>(getPathSegments());
-        segments.addAll(pathSegments.length > 0 ? Arrays.asList(pathSegments) : new ArrayList<String>());
-        setPathSegments(segments);
-        return this;
+        return appendPathSegments(Arrays.asList(pathSegments));
     }
 
     /**
--- a/httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java
@@ -507,7 +507,7 @@ public class URIBuilder {
     public URIBuilder appendPath(final String path) {
         final List<String> segments = new ArrayList<>(getPathSegments());
         segments.addAll(path != null ? splitPath(path) : new ArrayList<String>());
-        setPathSegments( segments );
+        setPathSegments(segments);
         return this;
     }
 
--- a/httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java
@@ -500,6 +500,18 @@ public class URIBuilder {
     }
 
     /**
+     * Appends path to URI. The value is expected to be unescaped and may contain non ASCII characters.
+     *
+     * @return this.
+     */
+    public URIBuilder appendPath(final String path) {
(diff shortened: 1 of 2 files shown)


ok2c:
@larshelge Looks good to me. Could you please add `#appendPathSegments` method for the sake of completeness?

larshelge:
Absolutely @ok2c. Added `appendPathSegments(String...)` and `appendPathSegments(List)`  now.

httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java
```diff
@@ -499,6 +499,18 @@ public URIBuilder setPath(final String path) {
...
 
+    /**
+     * Appends path to URI. The value is expected to be unescaped and may contain non ASCII characters.
+     *
+     * @return this.
+     */
+    public URIBuilder appendPath(final String path) {
+        final List<String> segments = new ArrayList<>(getPathSegments());
+        segments.addAll(path != null ? splitPath(path) : new ArrayList<String>());
+        setPathSegments(segments);
```
ok2c, line 510: @larshelge Should not it be possible to re-use `#setPathSegments` here?

larshelge: I fully agree. I thought about it but favored consistency with existing code, e.g. `setPathSegments(String...)` and `setPathSegments(List)`. I am happy to change it.

larshelge: Sorry, I did not fully understand - do you mean re-use `appendPathSegments(List)` ?

ok2c: > Sorry, I did not fully understand - do you mean re-use `appendPathSegments(List)` ?

@larshelge Yes, that is what I meant. I thought it would make it a bit neater. Not very important, though.

larshelge: @ok2c Got it. I will update it.

larshelge: @ok2c Made the changes now.

httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java
```diff
@@ -499,6 +499,16 @@ public URIBuilder setPath(final String path) {
         return this;
     }
 
+    /**
+     * Appends path to URI. The value is expected to be unescaped and may contain non ASCII characters.
+     *
+     * @return this.
+     */
+    public URIBuilder appendPath(final String path) {
```
ok2c, line 507: @larshelge One last thing. I suppose this should be a little more efficient by avoiding unnecessary operations and intermediate garbage.
```
    public URIBuilder appendPath(final String path) {
        if (path != null) {
             appendPathSegments(splitPath(path);
        }
        return this;
    }
```

httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java
```diff
@@ -541,6 +560,18 @@ public URIBuilder setPathSegments(final List<String> pathSegments) {
         return this;
     }
 
+    /**
+     * Appends segments to URI path. The value is expected to be unescaped and may contain non ASCII characters.
+     *
+     * @return this.
+     */
+    public URIBuilder appendPathSegments(final List<String> pathSegments) {
+        final List<String> segments = new ArrayList<>(getPathSegments());
```
ok2c, line 569: @larshelge Same here.

larshelge: Agreed @ok2c. I wanted to avoid if-statements but we can favor less heap usage here.

Note that the `new ArrayList<>(getPathSegments())` wrapper list is necessary as `getPathSegments` might return an immutable list which cannot be added to later.

I have made the updates now.

httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java
larshelge: Updated it now.

ok2c (approved):

PR #236, opened 2020-12-08
Add filters before MAIN_HANDLER so they won't be ignored

robtimus:
When `addFilterLast` is called on a `ServerBootstrap`, `AsyncServerBootstrap` or `H2ServerBootstrap`, the given filter is added as last element of the filter chain. That may seem correct, but it causes them to be ignored. The reason is the `Terminal*Filter` that's initially the last element. In all three cases, this does not delegate to the chain. As a result, anything added with `addFilterLast` is currently completely ignored.

This change adds the filter as last in the chain of actual filters, just before the `Terminal*Filter` of which it could be said it's not an actual filter.

### DIFF
--- a/httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/bootstrap/H2ServerBootstrap.java
+++ b/httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/bootstrap/H2ServerBootstrap.java
@@ -396,7 +396,9 @@ public class H2ServerBootstrap {
                         filterChainDefinition.addFirst(entry.filterHandler, entry.name);
                         break;
                     case LAST:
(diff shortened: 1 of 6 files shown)


carterkozak:
Thanks, @robtimus! Any chance you can include a test that fails without this fix?

robtimus:
@carterkozak: I added three new test classes:
* `ClassicServerBootstrapFilterTest`
* `AsyncServerBootstrapFilterTest`
* `H2ServerBootstrapFiltersTest`

I based these on `ClassicServerAndRequesterTest`, `Http1ServerAndRequesterTest` and `H2ServerAndRequesterTest` respectively, but stripped out most of the non-essential stuff. These fail because the `X-Test-Filter` header cannot be found when executed against master, and succeed when executed against these changes.

PR #238, opened 2020-12-23
HTTPCORE-656: Improve HttpEntity.getContent() documentation

Marcono1234:
See [HTTPCORE-656](https://issues.apache.org/jira/browse/HTTPCORE-656)

While writing this, I noticed that `HttpEntity` is also used by HTTP requests, therefore I added that this only applies to HTTP entities from HTTP responses.

Additionally the [tutorial](https://hc.apache.org/httpcomponents-client-ga/tutorial/html/fundamentals.html#d5e145) which mentions this behavior is part of the HTTP client project, therefore I decided against linking it from the documentation because that might be confusing.


ok2c:
@Marcono1234 Looks good to me. Could you please use the terms `incoming` and `outgoing` entity instead of response and request entity as HttpEtity is also used by the server transport components?

PR #242, opened 2021-01-04
fMinor Improvements:

arturobernalg:
* remove redundant initialization
* Add final

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/util/Deadline.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/util/Deadline.java
@@ -58,12 +58,12 @@ public class Deadline {
     /**
      * The maximum (longest-lived) deadline.
      */
-    public static Deadline MAX_VALUE = new Deadline(INTERNAL_MAX_VALUE);
+    public static final Deadline MAX_VALUE = new Deadline(INTERNAL_MAX_VALUE);
 
     /**
      * The minimum (shortest-lived) deadline.
      */
-    public static Deadline MIN_VALUE = new Deadline(INTERNAL_MIN_VALUE);
+    public static final Deadline MIN_VALUE = new Deadline(INTERNAL_MIN_VALUE);
 
     private static final SimpleDateFormat simpleDateFormat = new SimpleDateFormat(DATE_FORMAT);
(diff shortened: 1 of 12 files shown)


httpcore5/src/main/java/org/apache/hc/core5/util/Deadline.java
```diff
@@ -58,12 +58,12 @@
...
      * The maximum (longest-lived) deadline.
      */
-    public static Deadline MAX_VALUE = new Deadline(INTERNAL_MAX_VALUE);
+    public static final Deadline MAX_VALUE = new Deadline(INTERNAL_MAX_VALUE);
 
     /**
      * The minimum (shortest-lived) deadline.
      */
-    public static Deadline MIN_VALUE = new Deadline(INTERNAL_MIN_VALUE);
+    public static final Deadline MIN_VALUE = new Deadline(INTERNAL_MIN_VALUE);
```
ok2c, line 66: @arturobernalg Same problem as before. This will likely break binary compatibility with previous 5.x releases.

carterkozak: If memory serves, adding a final modifier doesn't break ABI unless something attempts to set the value (but I could be incorrect).
What if someone _is_ setting an unexpected value here? I think that would be much worse.

ok2c: @carterkozak I am pretty sure both Clirr and JApiCmd will mark this change as non-backward compatible. What one might want to do is to deprecate this value in favor of a final variable with a different name.

arturobernalg: reverted. Sorry @ok2c

carterkozak: @ok2c I think the analyzer is telling us that if we make the field final any users who already attempt to replace the default zero value with a different one will be broken, however it's my understanding that would result in other bugs because we expect MIN_VALUE to be zero (and constant) internally. The question is whether or not it's important to continue allowing library consumers to set the field with custom values.

PR #243, opened 2021-01-07
HTTPCORE-657: Add TrafficClass to IOReactorConfig

desmondyeung:


garydgregory:
Hi @desmondyeung 

Missing:
- unit tests
- Javadoc since tags for new public and protected methods.

desmondyeung:
@garydgregory I've added a unit test and the javadoc since tags.

PR #246, opened 2021-01-25
HTTPCORE-659 Race condition in IOSessionImpl when setting event

ncapponi:
I'm not including change on `setEventMask(final int newValue)` method as there is currently no lock and I have no knowledge on its usage. Tell me if you want this modification as well.
I'm not including a new test case because the race condition is quite hard to reproduce without some "artificial" help like making the code sleep in the `setEvent()` method.


ok2c:
@ncapponi Please do introduce locking to `#setEventMask`.

ncapponi:
@ok2c done

ncapponi:
Thanks !

PR #247, opened 2021-01-25
Minor Improvement:

arturobernalg:
* Use Empty collections
* Unnecessary boxing
* Unnecessary toString
* Use Math.max

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/http/ContentType.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/http/ContentType.java
@@ -105,6 +105,11 @@ public final class ContentType implements Serializable {
     public static final ContentType WILDCARD = create(
             "*/*", (Charset) null);
 
+    /**
+     * An empty immutable {@code NameValuePair} array.
+     */
+    private static final NameValuePair[] EMPTY_NAME_VALUE_PAIR_ARRAY = new NameValuePair[0];
+
 
     /**
      * @deprecated To be removed in 6.0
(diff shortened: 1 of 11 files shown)


httpcore5/src/main/java/org/apache/hc/core5/http/ContentType.java
```diff
@@ -105,6 +105,11 @@
     public static final ContentType WILDCARD = create(
             "*/*", (Charset) null);
 
+    /**
+     * An empty immutable {@code NameValuePair} array.
+     */
+    public static final NameValuePair[] EMPTY_NAME_VALUE_PAIR_ARRAY = new NameValuePair[0];
```
ok2c, line 111: @arturobernalg There is no need to make it public. Please change to private.

ok2c (changes requested):

httpcore5/src/main/java/org/apache/hc/core5/http/ContentType.java
arturobernalg: Ok. changed

arturobernalg (commented):
Changed

httpcore5/src/main/java/org/apache/hc/core5/http/impl/bootstrap/ServerBootstrap.java
```diff
@@ -407,7 +407,7 @@ public HttpServer create() {
         }
 
         return new HttpServer(
-                this.listenerPort > 0 ? this.listenerPort : 0,
+                Math.max(this.listenerPort, 0),
```
ok2c, line 410: @arturobernalg What are benefits of using `Math#max` here? Is it more byte code efficient? I am kind of old and stupid. It makes it more difficult for me to understand the intent of the operation. I am also a bit worried that the proposed code appears to make no distinction between greater and greater equal (as well lesser and lesser equal) comparison operations.

arturobernalg: > @arturobernalg What are benefits of using `Math#max` here? Is it more byte code efficient? I am kind of old and stupid. It makes it more difficult for me to understand the intent of the operation. I am also a bit worried that the proposed code appears to make no distinction between greater and greater equal (as well lesser and lesser equal) comparison operations.

HI @ok2c 
My idea was get more readability. As far as I know for implementation/performance sake, they would be nearly equivalent. Maybe a bit more faster the java implementations. But your right,  no make no distinction between greater and greater equal. I gonna remove that changes.

ok2c (approved):

PR #249, opened 2021-01-26
HTTPCORE-662 - add http status code 418

arturobernalg:
The RFC 2324 describes following HTTP Status codes:
418 I'm a teapot
Currently, the HTTP Status code list doesn't include this codes.


httpcore5/src/main/java/org/apache/hc/core5/http/HttpStatus.java
```diff
@@ -158,6 +159,8 @@ private HttpStatus() {
     public static final int SC_REQUESTED_RANGE_NOT_SATISFIABLE = 416;
     /** {@code 417 Expectation Failed} (HTTP/1.1 - RFC 7231) */
     public static final int SC_EXPECTATION_FAILED = 417;
+    /** {@code 418 I'm a teapot Failed} (HTTP/1.0 - RFC 2324) */
+    public static final int SC_I_AM_A_TEAPOT_FAILED = 418;
```
thc202, line 163: Was "Failed" added on purpose?

arturobernalg: Nop. Changed.

ok2c:
@arturobernalg Is RFC 2324 for real? Why do we need to support it?

arturobernalg:
> RFC 2324

Hi @ok2c 

at the time it seemed the correct option. Maybe i was wrong.

PR #257, opened 2021-02-10
Minor Improvements:

arturobernalg:
* Unnecessary boxing
* Unnecessary unboxing
* inline variable
* Change PosixParser to  DefaultParser


ok2c:
@arturobernalg Would it be OK for you to put javadoc changes into a separate commit or PR? I suspect there are many more places where javadocs need some work.

arturobernalg:
> @arturobernalg Would it be OK for you to put javadoc changes into a separate commit or PR? I suspect there are many more places where javadocs need some work.

Hi @ok2c . No problem. Do I need to create a Jira for the javadoc ?

ok2c:
@arturobernalg I would not bother. Just raise a PR when ready.

arturobernalg:
> @arturobernalg I would not bother. Just raise a PR when ready.

Hi @ok2c I've reverted the changes regarding the javaDoc. When i finish i'll provide another PR just only with javaDoc

TY

PR #258, opened 2021-02-11
Add / Fix / Complete javadoc

arturobernalg:
IMO we need a lot of work here with the javadoc. I'll try to continue a little bit every night. Please @ok2c  feel free to make any comment. I've some gap with mi English ;)

### DIFF
--- a/httpcore5-h2/src/main/java/org/apache/hc/core5/http2/nio/AsyncPingHandler.java
+++ b/httpcore5-h2/src/main/java/org/apache/hc/core5/http2/nio/AsyncPingHandler.java
@@ -49,6 +49,8 @@ public interface AsyncPingHandler {
      * Triggered to signal receipt of a ping response message.
      *
      * @param feedback the ping message feedback.
+     * @throws IOException in case of an I/O error
+     * @throws HttpException in case of HTTP protocol violation
      */
     void consumeResponse(ByteBuffer feedback) throws HttpException, IOException;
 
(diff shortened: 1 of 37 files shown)


httpcore5-h2/src/main/java/org/apache/hc/core5/http2/nio/AsyncPingHandler.java
```diff
@@ -49,6 +49,8 @@
      * Triggered to signal receipt of a ping response message.
      *
      * @param feedback the ping message feedback.
+     * @throws IOException in case of an I/O error
```
ok2c, line 52: @arturobernalg I honestly think this is just not worth it. `@throws IOException in case of an I/O error` adds no value to the javadocs. I am not sure what I am to do about this PR.

arturobernalg: hi @ok2c maybe you're right. Any advice to how proceed now? I'm honestly miss the javadoc in the above class and don't think it's important give all information that we can to the user's.

ok2c: @arturobernalg Are these all javadoc fixes you have in mind or are you planning on adding more at some point? Preferably I would like to avoid multiple `Add / Fix / Complete javadoc` commits.

arturobernalg: Hi @ok2c 

my idea it's progressively add the missing javadoc. I am not sure whether to do it in several PRs or in just one

ok2c: @arturobernalg To make my life a bit easier could you please keep those changes on a private batch and merge that branch with a single PR once ready?

arturobernalg: :) you're right. I apologize.

arturobernalg:
Close the  PR to make it private

PR #259, opened 2021-02-11
Add maven-bundle-plugin to create OSGi-ready packages

benapple:
The Felix maven-bundle-plugin adds a manifest file within the resulting jar that OSGi uses when wiring together dependencies. By default, the plugin ignores packages named impl, so explicit instructions have been added to the poms of httpcore5 and httpcore-h2 to publish all packages in the manifest. The resulting package is still a jar file, so non-OSGi consumers need not do anything different.


benapple:
Closing this pull request based on feedback from https://github.com/apache/httpcomponents-client/pull/284.

PR #260, opened 2021-02-12
HTTPCORE-667 - Add new Content Type application/problem+json

arturobernalg:
Add new content type --> application/problem+json
 
https://tools.ietf.org/html/rfc7807#section-6.1


michael-o (changes requested):
The code does not correspond to the Javadoc.

arturobernalg:
> The code does not correspond to the Javadoc.

True. Really sorry @michael-o.  huge mistake. Fixed

michael-o (approved):
LGTM.

PR #261, opened 2021-02-12
HTTPCORE-668 - W3C recommendation text/event-stream

arturobernalg:
Add new content type --> text/event-stream
Events W3C recommendation
https://www.w3.org/TR/eventsource/


ok2c:
@arturobernalg Do you have many of those still coming?

arturobernalg:
> @arturobernalg Do you have many of those still coming?

Hi @ok2c 

IMO we should give all the options to our users

ok2c:
@arturobernalg Fine by me. But please in a single PR and preferably without a JIRA for each and every instance on top it.

arturobernalg:
> @arturobernalg Fine by me. But please in a single PR and preferably without a JIRA for each and every instance on top it.

OK. agree. Just keep in mind that I'm looking at the code and when i see something that can help i create the Jira. I'll try to unify the changes

PR #265, opened 2021-02-18
5.2.x misc cleanups

garydgregory:
# Commit 1: Miscellaneous clean ups.
- Use isEmpty().
- Don't create empty arrays over and over.
- Don't nest else if you don't have to.
# Commit 2: Replace deprecated calls to Assert.assertThat.


ok2c:
@garydgregory Please port this change-set to master or wait until `5.2.x` has been properly re-based off `master`. First off, I have already removed `Assert#assertThat` in master. Secondly, you now have competion with those misc cleanups commits. There is no need to do that in each and every branch.

ok2c:
@garydgregory Please see 35873412ba799f78c953898e93e02e512d1a5437 in master.

garydgregory:
@ok2c 
Oops, I should have paid closer attention to the state of the branches. Sorry for the noise.

ok2c:
@garydgregory Please bear with me for a few more weeks. I would like to release 5.1 GA soon, retire 5.0 and promote 5.2 to master. Then we should be back to just one stable (5.1) and one development (5.2) branches, which is going to make it easier to keep track of things.

PR #267, opened 2021-02-25
HTTPCORE-670 -  Add new headers

arturobernalg:
Add new header

Set-Cookie --> https://tools.ietf.org/html/rfc2109#section-4.2.2
Link --> href="https://tools.ietf.org/html/rfc5988
Cookie --> https://tools.ietf.org/html/rfc2109
Content-Disposition --> https://tools.ietf.org/html/rfc6266
Access-Control-Request-Method --> https://www.w3.org/TR/cors
Access-Control-Request-Headers  --> https://www.w3.org/TR/cors
Access-Control-Max-Age --> https://www.w3.org/TR/cors
Access-Control-Expose-Headers --> https://www.w3.org/TR/cors
Access-Control-Allow-Origin --> https://www.w3.org/TR/cors
Access-Control-Allow-Methods --> https://www.w3.org/TR/cors
Access-Control-Allow-Credentials -->https://www.w3.org/TR/cors

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/http/HttpHeaders.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/http/HttpHeaders.java
@@ -48,6 +48,39 @@ public final class HttpHeaders {
 
     public static final String ACCEPT_RANGES = "Accept-Ranges";
 
+    /**
+     * The CORS {@code Access-Control-Allow-Credentials} response header field name.
+     */
+    public static final String ACCESS_CONTROL_ALLOW_CREDENTIALS = "Access-Control-Allow-Credentials";
+    /**
+     * The CORS {@code Access-Control-Allow-Headers} response header field name.
+     */
+    public static final String ACCESS_CONTROL_ALLOW_HEADERS = "Access-Control-Allow-Headers";
+    /**
+     * The CORS {@code Access-Control-Allow-Methods} response header field name.
+     */
+    public static final String ACCESS_CONTROL_ALLOW_METHODS = "Access-Control-Allow-Methods";
+    /**
+     * The CORS {@code Access-Control-Allow-Origin} response header field name.
+     */
+    public static final String ACCESS_CONTROL_ALLOW_ORIGIN = "Access-Control-Allow-Origin";
+    /**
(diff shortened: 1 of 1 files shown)


httpcore5/src/main/java/org/apache/hc/core5/http/HttpHeaders.java
```diff
@@ -130,6 +188,17 @@ private HttpHeaders() {
 
     public static final String STATUS_URI = "Status-URI";
 
+    /**
+     * The HTTP {@code Set-Cookie} header field name.
+     * @see <a href="https://tools.ietf.org/html/rfc2109#section-4.2.2">Section 4.2.2 of RFC 2109</a>
```
ok2c, line 193: @arturobernalg I am not sure about references to RFCs in the source code. For instance, HttpClient conforms to rfc6265 and not rfc2109. I would not want to scan the source code for RFC references every time protocol level gets upgraded.

httpcore5/src/main/java/org/apache/hc/core5/http/HttpHeaders.java
```diff
@@ -130,6 +188,17 @@ private HttpHeaders() {
...
+     */
+    public static final String SET_COOKIE = "Set-Cookie";
+    /**
+     * The HTTP {@code Set-Cookie2} header field name.
+     * @see <a href="https://tools.ietf.org/html/rfc2965">RFC 2965</a>
+     */
+    public static final String SET_COOKIE2 = "Set-Cookie2";
```
ok2c, line 200: @arturobernalg `Set-Cookie2` is considered obsolete and should not be used. I would not include it,

ok2c (changes requested):

httpcore5/src/main/java/org/apache/hc/core5/http/HttpHeaders.java
arturobernalg: Agree, changed

httpcore5/src/main/java/org/apache/hc/core5/http/HttpHeaders.java
arturobernalg: Agree, changed

ok2c (approved):

PR #275, opened 2021-03-27
Fix TextUtils.toHexString having incorrect result for negative values

Marcono1234:
The check `b < 16` was faulty and would have added a '0' for negative values.
E.g. for `(byte) 0xFF` the result would have been `"0ff"`.


ok2c:
@Marcono1234 Ported to `5.1.x` as well. Many thanks for contributing this fix.

PR #283, opened 2021-04-12
HTTPCLIENT-1916: Add URIBuilder.removeParameter

peterdettman:
https://issues.apache.org/jira/browse/HTTPCLIENT-1916 .

Note that I have marked the new method as "since 5.2".

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java
@@ -695,6 +695,27 @@ public class URIBuilder {
     }
 
     /**
+     * Removes parameter of URI query if set. The parameter name is expected to be unescaped and may
+     * contain non ASCII characters.
+     * <p>
+     * Please note query parameters and custom query component are mutually exclusive. This method
+     * will remove custom query if present, even when no parameter was actually removed.
+     * </p>
+     *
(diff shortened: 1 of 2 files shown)


garydgregory:
May you please test edge cases like null and ""?

peterdettman:
I'm happy to add such tests, but I'm a bit unclear what the expected behaviour ought to be. I'm inclined to require non-null, non-empty for `param` but actually other `URIBuilder` methods don't seem to exclude "" for a name (is that really permissible?). `BasicNameValuePair` guards only against null too).

peterdettman:
@michael-o Some rationale was given by the submitter; I am merely picking up a _volunteers-wanted_ label.

ok2c (approved):

ok2c:
> 
> 
> I'm happy to add such tests, but I'm a bit unclear what the expected behaviour ought to be. I'm inclined to require non-null, non-empty for `param` but actually other `URIBuilder` methods don't seem to exclude "" for a name (is that really permissible?). `BasicNameValuePair` guards only against null too).

@peterdettman I think just `null` check should be enough for now. The issue of name/value pairs with an empty name should be dealt with consistently across the entire library.

peterdettman:
I've added a null guard for the argument, with accompanying test.

michael-o (approved):

PR #284, opened 2021-04-21
Add a content length StringAsyncEntityProducer

imbyungjun:
Implement `StringAsyncEntityProducer::getContentLength` method

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/http/nio/entity/StringAsyncEntityProducer.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/http/nio/entity/StringAsyncEntityProducer.java
@@ -44,6 +44,7 @@ import org.apache.hc.core5.util.Asserts;
 public class StringAsyncEntityProducer extends AbstractCharAsyncEntityProducer {
 
     private final CharBuffer content;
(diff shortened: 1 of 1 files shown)


httpcore5/src/main/java/org/apache/hc/core5/http/nio/entity/StringAsyncEntityProducer.java
```diff
@@ -54,6 +55,7 @@ public StringAsyncEntityProducer(
         super(bufferSize, fragmentSizeHint, contentType);
         Args.notNull(content, "Content");
         this.content = CharBuffer.wrap(content);
+        this.length = content.toString().getBytes(contentType.getCharset()).length;
```
carterkozak, line 58: This is potentially very expensive in the case a large constant string is reused across many requests. Can we take some shortcuts, perhaps by validating the string is ascii, and using the input length in that case?

imbyungjun: I thought the implementation was missing, but it seems to be designed to produce HTTP requests without content-length header. Close this PR.

PR #285, opened 2021-05-18
Don't change conn flow control window based on remote SETTINGS

rschmitt:
From RFC7540 §6.9.2:

> The connection flow-control window can only be changed using
> WINDOW_UPDATE frames.


ok2c:
Also, according section to 6.9.2 Initial Flow-Control Window Size 
```
A SETTINGS frame cannot alter the connection flow-control window.
```
@rschmitt Good catch. Thank you.

ok2c (approved):

PR #288, opened 2021-06-08
Adds utility method to add name value pair for uri builder

anuragagarwal561994:
What?
Adding an overloaded method to uri builder to add NameValuePair as query param.

Why?
At times some of the query params are constant and do not change in terms of key and value.
In those cases it is better to make NameValuePair instances as static and re-using them instead of
creating an instance everytime.


ok2c:
@anuragagarwal561994 Could you please add `@since 5.2` tag?

anuragagarwal561994:
Done

PR #293, opened 2021-06-18
Fix data race in StrictConnPool

rschmitt:


httpcore5/src/main/java/org/apache/hc/core5/pool/StrictConnPool.java
```diff
@@ -699,14 +699,18 @@ public boolean isDone() {
         }
 
         public void failed(final Exception ex) {
-            if (this.completed.compareAndSet(false, true)) {
-                this.ex = ex;
+            synchronized (this.completed) {
```
carterkozak, line 702: Perhaps we replace the atomic boolean with an atomic reference to the result/exception, and avoid synchronization?

ok2c: @rschmitt Could you please give some details about the race? 
If we abolutely have to synchronize here the use of AtomicBoolean for that end feels wrong.

rschmitt: |Thread A | Thread B|
|-|-|
|Set `completed` to true ||
||Call `isDone()`, which returns `true`|
||Call `getResult()`/`getException()`, which return `null`|
|Set `ex`/`result`||

I found this race condition by inspection. I'm trying to track down a connection leak that we're continuing to get reports about. In the absence of a repro, I've resorted to auditing the code. I don't know if this race condition results in an actual bug, or is just benign, but either way there's no reason not to fix it.

carterkozak: I suppose an alternative and potentially simpler solution would be:
AtomicBoolean completed is used exclusively to guard internally _setting_ `this.result` and `this.completed`, but is no longer used by `public boolean isDone()`

isDone may be updated to make two volatile reads.
```java
public boolean isDone() {
   return result != null || ex != null;
}
```

This way isDone and getException/getResult always agree. What do you think?

ok2c: @carterkozak Makes sense to me.

carterkozak: I've created a PR with the proposed change here: #294

PR #295, opened 2021-07-07
Resolves HTTPCORE-682

veselov:
This change ads support for specifying JCA provider for loading trust manager/key managers.
Without this fix, the corresponding providers have to be loaded into the system which is not always desirable or possible.

I've refactored the `testBuildWithProvider..()` tests along the way, as they wouldn't have failed even if the `setProvider()` instruction was missing. The `DummyProvider` is quite wonky, but it does the job of verifying that the specified provider has been used.

LMK if there are any problems with this. The same change should go into 5.0/5.1 as well.

### DIFF
--- a/httpcore/src/test/java/org/apache/http/ssl/DummyProvider.java
+++ b/httpcore/src/test/java/org/apache/http/ssl/DummyProvider.java
@@ -1,0 +1,64 @@
+/*
+ * ====================================================================
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ * ====================================================================
+ *
+ * This software consists of voluntary contributions made by many
+ * individuals on behalf of the Apache Software Foundation.  For more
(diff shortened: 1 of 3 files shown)


ok2c:
@veselov The changeset looks good to me. However I do nto think it can go into the 4.4.x 5.0.x or 5.1.x stable branches due to introduction of new API methods. Could you please port it to master (5.2) first? Please also add `@since 5.2` to all new public methods.

ok2c (changes requested):
@veselov Please port to master.

httpcore/src/test/java/org/apache/http/ssl/DummyProvider.java
```diff
@@ -0,0 +1,64 @@
...
+ */
+
+package org.apache.http.ssl;
+
+import java.security.Provider;
+import java.security.Security;
+import java.util.HashSet;
+import java.util.Set;
+
+public class DummyProvider extends Provider {
```
ok2c, line 35: @veselov Should not it be possible to use Mockito spy for the same end here?

veselov: I've tried using `Mockito.spy` on returned `Service` objects to alter the `Service.getProvider()` response - that would have made it easier to test whether the provider was used. However, that was failing with some security exceptions. As to replacing the `DummyProvider` with a spied on class - I'm not sure how I can do that - i.e. what would I spy on, since the whole point of the change is for me to supply some alternative provider, so I need to use an implementation of `Provider`, and that's an abstract class.

Said that, I'm really not well versed on Mockito, if you have ideas how this can be simplified - I'll gladly adopt them.

veselov:
> @veselov The changeset looks good to me. However I do nto think it can go into the 4.4.x 5.0.x or 5.1.x stable branches due to introduction of new API methods. Could you please port it to master (5.2) first? Please also add `@since 5.2` to all new public methods.

Changes for master: https://github.com/apache/httpcomponents-core/pull/296

However, I do see that some APIs are added in the patch releases. For example `SSLContextBuilder.setKeyStoreType()` has been added in `4.4.7`. Has the policy been changed since then?

ok2c:
> Has the policy been changed since then?
@veselov Yes, it has. Things got really out of hand during the 5.0 ALPHA and BETA development phases for various reasons, resource shortage being the main one.

ok2c:
Superseded by #296

PR #296, opened 2021-07-11
HTTPCORE-682: Custom provider for key manager/trust manager initialization

veselov:
This change ads support for specifying JCA provider for loading trust manager/key managers.
Without this fix, the corresponding providers have to be loaded into the system which is not always desirable or possible.

I've refactored the `testBuildWithProvider..()` tests along the way, as they wouldn't have failed even if the `setProvider()` instruction was missing. The `DummyProvider` is quite wonky, but it does the job of verifying that the specified provider has been used.

### DIFF
--- a/httpcore5/src/test/java/org/apache/hc/core5/ssl/TestSSLContextBuilder.java
+++ b/httpcore5/src/test/java/org/apache/hc/core5/ssl/TestSSLContextBuilder.java
@@ -37,6 +37,7 @@ import java.net.URL;
 import java.security.KeyStore;
 import java.security.KeyStoreException;
 import java.security.NoSuchAlgorithmException;
+import java.security.NoSuchProviderException;
 import java.security.Principal;
 import java.security.Security;
 import java.security.UnrecoverableKeyException;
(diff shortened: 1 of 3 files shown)


httpcore5/src/test/java/org/apache/hc/core5/ssl/TestSSLContextBuilder.java
```diff
@@ -27,6 +27,19 @@
...
 package org.apache.hc.core5.ssl;
 
+import org.apache.hc.core5.util.Timeout;
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Test;
+
+import javax.net.ssl.KeyManagerFactory;
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLException;
```
ok2c, line 37: @veselov Please avoid moving imports around. It just causes unnecessary noice in the chanage-set. Looks good otherwise.

ok2c (changes requested):

httpcore5/src/test/java/org/apache/hc/core5/ssl/TestSSLContextBuilder.java
veselov: Done.

PR #297, opened 2021-08-11
Refactor IO reactor close timeout.

garydgregory:
5.1.x refactor reactor close timeout into a new method that accepts a timeout.


ok2c:
@garydgregory All API visible changes should go to `master`. I see no reason why this change needs to be in 5.1

garydgregory:
> @garydgregory All API visible changes should go to `master`. I see no reason why this change needs to be in 5.1

Closing per the above from Oleg.

PR #303, opened 2021-10-15
Let ResourceHolder implement java.lang.AutoCloseable.

garydgregory:
- This happens to highlight where tests using implementations of
ResourceHolder where not testing closing down any resources. Tests are
updated but only close the resource, assertions to validate close states
are not added.
- JApiCmp does not seem to know that Java 8's default methods provide
binary and source compatibility.


ok2c:
@garydgregory I am _strongly_ against this change. Async data consumers and producers _intentionally_ do not implement `AutoCloseable` as they cannot be used with try-with-resources. Please do not merge.

PR #310, opened 2021-10-28
5.2.x java time session and deadline

garydgregory:
Use Java 8's java.time package in sessions and deadlines.


michael-o (commented):
I agree with @ok2c `Instant` gives here zero benefit compared with `long`.

PR #311, opened 2021-10-30
HTTPCORE-690   -Implement H2StreamMultiplexerFactory fluent builders

arturobernalg:


ok2c:
@arturobernalg These are internal classes not intended for use outside of the library. Do they _really_ need a builder?

arturobernalg:
> @arturobernalg These are internal classes not intended for use outside of the library. Do they _really_ need a builder?

HI @ok2c probably not. I close the PR.
TY

PR #317, opened 2021-11-04
Remove a check if inPlainBuf has remaining capacity prior to writing …

JasonMathison:
…it out.  This causes an endless loop if the inEncrypted buffer is larger than the inPlainBuf.

https://issues.apache.org/jira/browse/HTTPCORE-694

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/reactor/ssl/SSLIOSession.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/reactor/ssl/SSLIOSession.java
@@ -588,7 +588,6 @@ public class SSLIOSession implements IOSession {
                         if (sslEngine.isInboundDone()) {
                             endOfStream = true;
                         }
-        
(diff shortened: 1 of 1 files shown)


httpcore5/src/main/java/org/apache/hc/core5/reactor/ssl/SSLIOSession.java
```diff
@@ -588,14 +588,14 @@ private void decryptData(final IOSession protocolSession) throws IOException {
                         if (sslEngine.isInboundDone()) {
                             endOfStream = true;
                         }
-                        if (inPlainBuf.hasRemaining()) {
```
ok2c, line 591: @JasonMathison Oh, damn. This check should have been `inPlainBuf.position() > 0`. The idea is to avoid doing the buffer flip and a useless input event with an empty buffer. Could you please test code with this change and let me know if everything still works as expected?

PR #325, opened 2021-11-11
Use cameCase in isSoKeepAlive method.

arturobernalg:
Probably it's good idea make this change now that we are in a new Release.


ok2c (changes requested):
@arturobernalg This change breaks API compatibility and breaks the build. Please make sure the old method is kept and deprecated.

arturobernalg:
> @arturobernalg This change breaks API compatibility and breaks the build. Please make sure the old method is kept and deprecated.

HI @ok2c
Deprecated.
TY

ok2c (approved):

PR #326, opened 2021-11-11
Fix typo, better name and use camelCase.

arturobernalg:


ok2c:
@arturobernalg I understand why you want to do that but this will likely make merging changes between branches (5.1 -> master) more difficult. It is not worth it in my opinion.

arturobernalg:
> @arturobernalg I understand why you want to do that but this will likely make merging changes between branches (5.1 -> master) more difficult. It is not worth it in my opinion.

Fair enough  @ok2c.

PR #327, opened 2021-11-12
HTTPCORE-699 - Use Instant.now() across httpComponents.

arturobernalg:


ok2c:
@arturobernalg And what is exactly the benefit of doing `Instant.now().toEpochMilli()` over `System.currentTimeMillis()`? Generating pointless garbage on the heap? 
Please do _not_ do that.

arturobernalg:
> @arturobernalg And what is exactly the benefit of doing `Instant.now().toEpochMilli()` over `System.currentTimeMillis()`? Generating pointless garbage on the heap?
  > Please do _not_ do that.
  
  OK

garydgregory:
FYI, Instants are NOT allocated in the heap, as it is a value-type.

carterkozak:
@garydgregory Do you have documentation on that? I'd thought the allocations in hot loops could often be elided, but I wasn't aware that was always the case.

garydgregory:
The Javadoc says it's a value type so instances are data bytes. Granted a value based intrance within an object that is itself in the heap is also in the heap.
https://docs.oracle.com/javase/8/docs/api/java/time/Instant.html
https://docs.oracle.com/javase/8/docs/api/java/lang/doc-files/ValueBased.html
https://openjdk.java.net/jeps/390
I can't find the pdf or blog on oracle.com that talks more about this.

ok2c:
> FYI, Instants are NOT allocated in the heap, as it is a value-type.
@garydgregory No, they are NOT. Value types are NOT supported and will not be supported until the project Valhalla has been complete [1]. Presently this is _merely_ a hint that one day `Instant` instances and the likes _might_ be allocated on the stack rather than on the heap in the future.

[1] https://openjdk.java.net/projects/valhalla/

garydgregory:
Well, ok, then. Ok2c's point is still valid that it does not make sense to create an Instant only to get its epoch millis of course. If we keep data as an instant instant instead of a long, that is fine IMO.

PR #330, opened 2021-11-14
Use Args for argument check.

arturobernalg:


ok2c:
@arturobernalg For love of all that is holy to you, please channel your time and effort into something more useful. Why would you want to replace a simple `not null` check with a utility method that does a whole bunch of `instanceof` checks internally? Why? HttpCore is a low level library. It needs to be reasonably memory and execution efficient. All those unnecessary checks and heap garbage add up.

arturobernalg:
OK

PR #333, opened 2021-11-20
HTTPCORE-700 - HttpHost genetic constructor NamedEndpoint

arturobernalg:
Allow create an instance of HttpHost with an any implementations of NamedEndpoint.  Now we are able create HttpHost with the next implementations: 
```
* Host
* URIAuthority --> already implemented.
* org.apache.hc.core5.net.NamedEndpoint
```

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/http/HttpHost.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/http/HttpHost.java
@@ -37,6 +37,7 @@ import org.apache.hc.core5.annotation.ThreadingBehavior;
 import org.apache.hc.core5.net.Host;
 import org.apache.hc.core5.net.NamedEndpoint;
 import org.apache.hc.core5.net.URIAuthority;
+import org.apache.hc.core5.reactor.EndpointParameters;
 import org.apache.hc.core5.util.Args;
 import org.apache.hc.core5.util.LangUtils;
 import org.apache.hc.core5.util.TextUtils;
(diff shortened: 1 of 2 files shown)


httpcore5/src/main/java/org/apache/hc/core5/http/HttpHost.java
```diff
@@ -255,6 +255,22 @@ public HttpHost(final URIAuthority authority) {
...
+    /**
+     * Creates {@code HttpHost} instance with the default scheme and the given {@link
+     * NamedEndpoint}
+     *
+     * @throws IllegalArgumentException If the port parameter is outside the specified range of
+     *                                  valid port values, which is between 0 and 65535, inclusive.
+     *                                  {@code -1} indicates the scheme default port.
+     * @see Host
+     * @see URIAuthority
+     * @see org.apache.hc.core5.reactor.EndpointParameters
```
ok2c, line 267: @arturobernalg This looks wrong

httpcore5/src/test/java/org/apache/hc/core5/http/TestHttpHost.java
```diff
@@ -69,6 +73,27 @@ public void testConstructor() {
         Assert.assertThrows(NullPointerException.class, () -> new HttpHost(null, (String) null, -1));
         Assert.assertThrows(IllegalArgumentException.class, () -> new HttpHost(null, "   ", -1));
         Assert.assertThrows(NullPointerException.class, () -> new HttpHost(null, (InetAddress) null, -1));
+
+        final Host host = new Host("myHost", 8080);
```
ok2c, line 77: @arturobernalg Please add a test case with non-default scheme such as `https`. I suspect there will be an unpleasant surprise.

ok2c (changes requested):

httpcore5/src/test/java/org/apache/hc/core5/http/TestHttpHost.java
arturobernalg: HI @ok2c 
I know. but it the same  whit ->
`public HttpHost(final URIAuthority authority)`
What you suggest that we should do?

ok2c: > I know. but it the same whit the Authority constructor

@arturobernalg It is not the same. The URI authority does not have a scheme component.
> What you suggest that we should do?

I suggest that we do nothing. Not only this constructor is not very useful, it can lead to incorrect / unpredictable results.

ok2c: No, not nothing. I am going to deprecate `public HttpHost(final URIAuthority authority)`.

arturobernalg: OK. 
Close the PR:
TY

ok2c:
@arturobernalg Please also close the associated JIRA.

arturobernalg:
> @arturobernalg Please also close the associated JIRA.

done.

PR #339, opened 2021-12-15
Logs improvements:

arturobernalg:
* Use just log4j-api.
* Rename libs to log4j2.
* Bump version slf4j 1.7.32.


arturobernalg:
@michael-o 
idem here pls.
TY

PR #340, opened 2021-12-22
Add iterator with Cache Control http header as a default.

arturobernalg:


ok2c:
@arturobernalg What is this good for? There is already a generic `#iterate` method.

arturobernalg:
> @arturobernalg What is this good for? There is already a generic `#iterate` method.

to replace e.g 
https://github.com/apache/httpcomponents-client/blob/master/httpclient5-cache/src/main/java/org/apache/hc/client5/http/impl/cache/CachedResponseSuitabilityChecker.java

ok2c:
@arturobernalg Please use the generic `#iterate` instead.

PR #342, opened 2021-12-30
Use status equality enums.

arturobernalg:
easy to read.


ok2c:
@arturobernalg Useless.

PR #345, opened 2022-03-03
BasicHttpRequest getUri() not reflecting setUri(); fragment not taken into account in setUri()

antpyx:
Fix BasicHttpRequest getUri() not reflecting setUri() #1; fix fragment not taken into account in setUri() #2


ok2c:
@antpyx This is per design and is in conformance with the HTTP spec. URI fragments can be used in HTML pages for navigation but are not to be included in request URIs.

PR #346, opened 2022-03-21
Upgrade to RxJava3

rschmitt:
Since the `ReactiveTestUtils` API references symbols from RxJava2, I've
left it untouched for now so that japicmp doesn't yell at me.

### DIFF
--- a/httpcore5-reactive/pom.xml
+++ b/httpcore5-reactive/pom.xml
@@ -55,9 +55,9 @@
       <scope>test</scope>
     </dependency>
     <dependency>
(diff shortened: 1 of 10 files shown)


ok2c:
> Since the ReactiveTestUtils API references symbols from RxJava2, I've left it untouched for now so that japicmp doesn't yell at me.

@rschmitt Sounds good to me. You might want to mark `Reactive3TestUtils ` as internal, if it is not meant to be a part of the public APIs.

ok2c (approved):

PR #351, opened 2022-06-22
HTTPCLIENT-2222: fixed ProtocolNegotiationException

gaojieliu:
Aligned with the implementation of the latest release to get rid of
ProtocolNegotiationException:
https://github.com/apache/httpcomponents-core/blob/rel/v5.2-beta2/httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/ClientH2PrefaceHandler.java#L142

cc @ok2c

### DIFF
--- a/httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/H2OnlyClientProtocolNegotiator.java
+++ b/httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/H2OnlyClientProtocolNegotiator.java
@@ -36,6 +36,7 @@ import org.apache.hc.core5.annotation.Internal;
 import org.apache.hc.core5.concurrent.FutureCallback;
 import org.apache.hc.core5.http.HttpVersion;
 import org.apache.hc.core5.http.ProtocolVersion;
+import org.apache.hc.core5.http.impl.nio.BufferedData;
 import org.apache.hc.core5.http2.ssl.ApplicationProtocol;
 import org.apache.hc.core5.reactor.IOSession;
 import org.apache.hc.core5.reactor.ProtocolIOSession;
@@ -58,6 +59,7 @@ public class H2OnlyClientProtocolNegotiator extends ProtocolNegotiatorBase {
     private final AtomicBoolean initialized;
 
     private volatile ByteBuffer preface;
+    private volatile BufferedData inBuf;
 
     public H2OnlyClientProtocolNegotiator(
             final ProtocolIOSession ioSession,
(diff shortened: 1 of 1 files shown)


httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/H2OnlyClientProtocolNegotiator.java
```diff
@@ -104,7 +106,8 @@ private void writeOutPreface(final IOSession session) throws IOException  {
         }
         if (!preface.hasRemaining()) {
             session.clearEvent(SelectionKey.OP_WRITE);
-            startProtocol(HttpVersion.HTTP_2, new ClientH2IOEventHandler(http2StreamHandlerFactory.create(ioSession)), null);
+            final ByteBuffer data = inBuf != null ? inBuf.data() : null;
```
carterkozak, line 109: We can avoid the unnecessary additional volatile read (I find this pattern helpful in general, especially when null checks are involved, though I realize in this case we never transition from non-null to null)
```suggestion
            final BufferedData inBufSnapshot = inBuf;
            final ByteBuffer data = inBufSnapshot != null ? inBufSnapshot.data() : null;
```

httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/H2OnlyClientProtocolNegotiator.java
```diff
@@ -131,7 +134,10 @@ public void outputReady(final IOSession session) throws IOException {
     @Override
     public void inputReady(final IOSession session, final ByteBuffer src) throws IOException {
         if (src != null) {
-            throw new ProtocolNegotiationException("Unexpected input");
+            if (inBuf == null) {
+                inBuf = BufferedData.allocate(src.remaining());
+            }
+            inBuf.put(src);
```
carterkozak, line 140: I think we need to handle a potential buffer overflow here in the case an existing `inBuf` doesn't have enough room for `src`

httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/H2OnlyClientProtocolNegotiator.java
ok2c: @carterkozak `IOEventHandler`s can be called by a single I/O event dispatch thread at a time by design and therefore do not need to synchronize access to their internal structures if they do not delegate event handling to external handlers.

ok2c:
@gaojieliu These are reasonable changes, though, I thought they were only relevant to protocol negotiators in `master` that have been completely rewritten in the course of 5.2 development.

gaojieliu:
Hi @ok2c 
Thanks for the update, and enjoy your vacation!

gaojieliu:
Hi @garydgregory 
Thank a lot, and let me give it a try!

PR #358, opened 2022-08-01
Fix unrepeatable producers to return isRepeatable false

HeartPattern:
These classes are returning true for `isRepeatable` but actually not repeatable.
They does not reset bytebuf pointer after producing, and some of them clear bytebuf on `releaseResources`, which make producer produce empty data on repeat.


ok2c:
@HeartPattern Would not fixing the resource management logic in those producers be a better approach?

ok2c:
Your assertion about `BasicAsyncEntityProducer` and `StringAsyncEntityProducer` producers being unrepeatable is wrong. They are. There are even test cases that verify that particular functionality. 

As far as `Buffer#clear` is concerned see Buffer Javadocs:
```
This method does not actually erase the data in the buffer, but it is named as if it did because it will most often be used in situations in which that might as well be the case.
```

PR #373, opened 2022-11-01
Make it possible to read chunked entities with maxResultLength

andersjaensson:
This makes it possible to read more than 4096 bytes of a `HttpEntity` with `EntityUtils.toByteArray` when you provide a `maxResultLength`, even if the entity doesn't have a content-length set (for example chunked).

It now reads the stream in a simliar way as `toCharArrayBuffer`.


andersjaensson:
@ok2c thanks for merging this, do you think it makes sense to backport to 5.1?

ok2c:
@andersjaensson 5.2.x is about to go GA and 5.1.x is about to get discontinued. I would not bother with a backport unless you insist.

andersjaensson:
@ok2c I would love to have it in a 5.1 release as soon as possible, since that is what we're using in production, but I'll leave that decision to you :)
When do you think there will be a 5.2.x GA?

ok2c:
@andersjaensson I back-ported your changes to 5.1.x. Please note, though. 5.1 release will take place _after_ the 5.2 GA

PR #385, opened 2023-02-05
Replace lambdas with method references.

arturobernalg:
Method references are more clear and readable compared to. Of course this  could be a matter of taste.


michael-o (approved):
For me, it reads nicer.

PR #388, opened 2023-02-10
Fix the issue with invalid scoped IPv6 addresses in InetAddressUtils.

arturobernalg:
This Pull Request addresses the issue of properly handling scoped IPv6 addresses in the InetAddressUtils class. The previous implementation had incorrect behavior when handling scoped IPv6 addresses. This fix updates the logic to properly check the format of the scope ID and to return false when the address is not a valid standard or compressed IPv6 address.

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/net/InetAddressUtils.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/net/InetAddressUtils.java
@@ -79,6 +79,19 @@ public class InetAddressUtils {
                  "::" +
                  "(([0-9A-Fa-f]{1,4}(:[0-9A-Fa-f]{1,4}){0,5})?)$"); // 0-6 hex fields
 
+    /**
+     * Regular expression pattern to match the scope ID in an IPv6 scoped address.
+     * The scope ID should be a non-empty string consisting of only alphanumeric characters or "-".
+     */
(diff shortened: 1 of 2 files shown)


httpcore5/src/main/java/org/apache/hc/core5/net/InetAddressUtils.java
```diff
@@ -139,7 +152,32 @@ public static boolean isIPv6HexCompressedAddress(final String input) {
...
+
+    /**
+     * Determines whether the given string represents a valid scoped IPv6 address.
+     * A scoped IPv6 address includes a scope identifier (e.g. "fe80::1ff:fe23:4567:890a%eth2").
+     *
+     * @param input the string to be tested
+     * @return {@code true} if the string represents a valid scoped IPv6 address, {@code false} otherwise
+     */
+    private static boolean isIPv6ScopedAddress(final String input) {
+        final int index = input.indexOf(SCOPE_ID_DELIMITER);
```
ok2c, line 171: @arturobernalg How about folding this method into `#isIPv6Address` and doing `input.indexOf(SCOPE_ID_DELIMITER)` only once?

ok2c (changes requested):

ok2c:
@arturobernalg Could you please change the target branch to 5.3.x? Otherwise looks good.

PR #390, opened 2023-02-12
Add a "Forwarded" HttpRequestInterceptor to add the "Forwarded" HTTP header.

arturobernalg:
This pull request adds support for the Forwarded header to improve debugging, statistics, and location-dependent content generation. The Forwarded header contains information added by reverse proxy servers (load balancers, CDNs, etc.) that would otherwise be altered or lost when proxy servers are involved in the request path. The header can be used to identify the original request's IP address, host, and protocol, which can be valuable information for debugging, tracking traffic, and generating location-specific content.

By adding support for the Forwarded header, we can ensure that the complete request information is captured and available for use in debugging, tracking, and content generation. This can lead to more accurate logs, improved site performance, and better user experiences.

Usage

To use this class, simply call ForwardedRequest.INSTANCE in the addInterceptor method of the HttpClient object that you want to use. For example:

```
CloseableHttpClient httpClient = HttpClientBuilder.create()
    .addInterceptorFirst(ForwardedRequest.INSTANCE)
    .build();
```

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/http/protocol/ForwardedRequest.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/http/protocol/ForwardedRequest.java
@@ -1,0 +1,186 @@
+/*
+ * ====================================================================
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ * ====================================================================
+ *
+ * This software consists of voluntary contributions made by many
+ * individuals on behalf of the Apache Software Foundation.  For more
+ * information on the Apache Software Foundation, please see
+ * <http://www.apache.org/>.
+ *
+ */
+
+package org.apache.hc.core5.http.protocol;
+
+import java.io.IOException;
+import java.net.Inet6Address;
+import java.net.InetSocketAddress;
+import java.net.SocketAddress;
+
+import org.apache.hc.core5.annotation.Contract;
+import org.apache.hc.core5.annotation.ThreadingBehavior;
+import org.apache.hc.core5.http.EndpointDetails;
+import org.apache.hc.core5.http.EntityDetails;
+import org.apache.hc.core5.http.Header;
+import org.apache.hc.core5.http.HttpException;
+import org.apache.hc.core5.http.HttpRequest;
+import org.apache.hc.core5.http.HttpRequestInterceptor;
+import org.apache.hc.core5.http.HttpVersion;
+import org.apache.hc.core5.http.ProtocolException;
+import org.apache.hc.core5.http.ProtocolVersion;
+import org.apache.hc.core5.net.URIAuthority;
+import org.apache.hc.core5.util.Args;
+
+/**
+ * The ForwardedRequest class is an implementation of the {@link HttpRequestInterceptor} interface
+ * that can be used by a proxy server to add a Forwarded header to an HTTP request. The Forwarded
+ * header is used to capture information about the intermediate nodes that a request has passed
+ * through. This information can be useful for security purposes or for debugging purposes.
+ * <p>
+ * The Forwarded header consists of a list of key-value pairs separated by semicolons. The keys that
+ * can be used in the Forwarded header include "host", "port", "proto", "for", and "by". The host
+ * key is used to specify the host name or IP address of the request authority. The port key is used
+ * to specify the port number of the request authority. The proto key is used to specify the
+ * protocol version of the request. The for key is used to specify the IP address of the client
+ * making the request. The by key is used to specify the IP address of the node adding the Forwarded
+ * header.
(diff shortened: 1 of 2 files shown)


httpcore5/src/main/java/org/apache/hc/core5/http/protocol/ForwardedRequest.java
```diff
@@ -0,0 +1,159 @@
...
+        }
+
+        final String byValue = (String) context.getAttribute(X_FORWARDED_HEADER_NAME);
+        if (byValue != null) {
+            valueBuilder.append(";by=").append(byValue);
+        } else {
+            valueBuilder.append(";by=unknown");
+        }
+
+        final Header header = request.getFirstHeader(X_FORWARDED_HEADER_NAME);
```
ok2c, line 151: @arturobernalg If all you want is to override a header value if it has already been set, use `Header#setHeader`. It is massively more efficient then `Header#getFirstHeader`, `Header#removeHeader`, `Header#addHeader` sequence.

arturobernalg: change it.

ok2c: @arturobernalg Why do you still need `Header#getFirstHeader` and `Header#removeHeader`?

ok2c (changes requested):

ok2c (approved):

ok2c:
@arturobernalg One more thing. The javadocs say
```
 The Forwarded header is added to the request by calling the `HttpRequest.addHeader` method. If a
 Forwarded header already exists in the request, the new key-value pairs are appended to the
 existing header.
 ```
but this is not what the implementation currently does and I think it did not do that before, either. Is the interceptor meant to preserve the list of previous hops in the forwarded chain or it is meant to build the initial list consisting of the first hop only? Please clarify.

**Comment by arturobernalg** [comment]
> @arturobernalg One more thing. The javadocs say
> 
> ```
>  The Forwarded header is added to the request by calling the `HttpRequest.addHeader` method. If a
>  Forwarded header already exists in the request, the new key-value pairs are appended to the
>  existing header.
> ```
> 
> but this is not what the implementation currently does and I think it did not do that before, either. Is the interceptor meant to preserve the list of previous hops in the forwarded chain or it is meant to build the initial list consisting of the first hop only? Please clarify.

@ok2c my mistake. the existing header is overwritten with the newly generated value

**Comment by ok2c** [comment]
@arturobernalg I am still not entirely sure I fully understand the intent of this interceptor. Is it meant to be used by (1) the client to populate the forwarded value for the very hop, by (2) the proxy to add one more hop to the existing list of hops, or by (3) both. At the moment the interceptor appears to be intended for the scenario 1 only, but in such a case I think it should not meddle if the `Forwarded` header if it has already been set by the caller.

**Comment by arturobernalg** [comment]
I did the necessary change to be complies to the RFC.

**Review by ok2c — CHANGES_REQUESTED** [review]
@arturobernalg Logically the interceptor now behaves correctly, but let's optimize the code a little.

File: httpcore5/src/main/java/org/apache/hc/core5/http/protocol/ForwardedRequest.java [hunk 2]
```diff
@@ -0,0 +1,189 @@
...
+
+        final URIAuthority authority = request.getAuthority();
+        if (authority == null) {
+            throw new ProtocolException("Request authority not specified");
+        }
+
+        final int port = authority.getPort();
+
+        final StringBuilder valueBuilder = new StringBuilder();
+        final String forwardedHeaderValue = request.getFirstHeader(FORWARDED_HEADER_NAME) != null ?
```
ok2c, line 128: @arturobernalg Maybe the compiler can optimize it, but I doubt it. So help it. Use a variable to store the result of `request.getFirstHeader(FORWARDED_HEADER_NAME)`

httpcore5/src/main/java/org/apache/hc/core5/http/protocol/ForwardedRequest.java
```diff
@@ -0,0 +1,189 @@
...
+                if (inetAddress.getAddress() instanceof Inet6Address) {
+                    valueBuilder.append("by=\"").append(byValue).append("\"");
+                } else {
+                    valueBuilder.append("by=").append(byValue);
+                }
+            }
+        }
+
+        // Add the "for" parameter
+        if (endpointDetails != null) {
```
ok2c, line 155: @arturobernalg It should be enough to do the `if (endpointDetails != null)` once.

httpcore5/src/main/java/org/apache/hc/core5/http/protocol/ForwardedRequest.java
```diff
@@ -0,0 +1,189 @@
...
+                if (inetAddress.getAddress() instanceof Inet6Address) {
+                    valueBuilder.append(";for=\"").append(forValue).append("\"");
+                } else {
+                    valueBuilder.append(";for=").append(forValue);
+                }
+            }
+        }
+
+        // Add the "host" parameter
+        final String hostValue = authority.getHostName();
```
ok2c, line 169: @arturobernalg You can safely assume `#getHostName` of the authority is never `null`.

ok2c (changes requested):

arturobernalg:
HI @ok2c 
I just commit all the change that you suggested.
TY

ok2c:
@arturobernalg Almost there. Now consider the case when the `EndpointDetails` is null (not present in the execution context). What `forwarded` value would the interceptor produce? Please make sure that your code either adds a valid header, throws an exception or does nothing. It should never produce a malformed header no matter circumstances. A test case for the `EndpointDetails` being null would also be nice.

arturobernalg:
> @arturobernalg Almost there. Now consider the case when the `EndpointDetails` is null (not present in the execution context). What `forwarded` value would the interceptor produce? Please make sure that your code either adds a valid header, throws an exception or does nothing. It should never produce a malformed header no matter circumstances. A test case for the `EndpointDetails` being null would also be nice.

@ok2c  done

PR #392, opened 2023-02-23
Hygiene changes to Deadline

jkmcl:
Since this is the v5.3 branch which is in alpha stage, is it fine to break binary compatibility by making Deadline.MIN_VALUE and Deadline.MAX_VALUE final and removing the checked exception from Deadline.parse since it cannot be thrown from the method body?

For context regarding MIN_VALUE/MAX_VALUE: https://github.com/apache/httpcomponents-core/pull/355

Thanks.


ok2c:
@jkmcl This cannot be done until 6.0.

PR #394, opened 2023-02-24
Add support for Internationalized Domain Names (IDN)

arturobernalg:
Improve IDN support in URI parsing by converting non-ASCII characters to their ASCII-compatible form using the java.net.IDN class. This allows for better interoperability with systems that may not support Unicode characters in domain names. The change was made by adding a conditional statement that checks if the hostname contains non-ASCII characters, and if so, applies the IDN conversion.

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/util/TextUtils.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/util/TextUtils.java
@@ -141,4 +141,22 @@ public final class TextUtils {
         return s.toLowerCase(Locale.ROOT);
     }
 
+
+    /**
+     * Determines whether the given {@link CharSequence} contains only ASCII characters.
+     *
+     * @param s the {@link CharSequence} to check
+     * @return true if the {@link CharSequence} contains only ASCII characters, false otherwise
+     * @throws IllegalArgumentException if the input {@link CharSequence} is null
+     * @since 5.3
+     */
+    public static boolean isAllASCII(final CharSequence s) {
+        for (int i = 0; i < s.length(); i++) {
+            if (s.charAt(i) > 0x7F) {
+                return false;
+            }
+        }
+        return true;
+    }
+
 }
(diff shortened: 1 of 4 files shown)


httpcore5/src/main/java/org/apache/hc/core5/util/TextUtils.java
```diff
@@ -141,4 +141,22 @@ public static String toLowerCase(final String s) {
...
+
+    /**
+     * Determines whether the given string contains only ASCII characters.
+     *
+     * @param s the string to check
+     * @return true if the string contains only ASCII characters, false otherwise
+     * @throws IllegalArgumentException if the input string is null
+     * @since 5.3
+     */
+    public static boolean isAllASCII(final String s) {
```
ok2c, line 153: @arturobernalg Please make the new method use `CharSequence` as a parameter instead of `String`.

ok2c (changes requested):
@arturobernalg I suppose this change-set should go into `5.3.x`, not `master`. Looks good otherwise.

httpcore5/src/main/java/org/apache/hc/core5/util/TextUtils.java
arturobernalg: done.

arturobernalg:
I just change the branch base to 5.3.x and use CharSequence instead

ok2c:
@arturobernalg I have to say I did not do a good job reviewing your original change-set. With your changes one could create instances of `Host` class referring to the same IDN host but with and without IDN encoding. That was wrong and could potentially lead to various issues with IDN hostnames. 

I changed the internal representation of the hostname to Unicode and added IDN conversion to those routines that parse hostnames from text or format them back. This way the internal representation is always consistent regardless of the original representation or formatting.

I re-committed your original change-set as 716f64c744d227bca80d8ee6a0a5cc951a5910d9. Please review. This should also make apache/httpcomponents-client#431 unnecessary.

arturobernalg:
@ok2c 
I need to digest everything, but it seems that you are correct.

PR #396, opened 2023-03-19
[HTTPCORE-736]HttpClient GET method adds Content-Length: 0 header

XenoAmess:
https://issues.apache.org/jira/browse/HTTPCORE-736

HttpClient GET method adds Content-Length: 0 header
This would cuase some server/firewall deny this kind of http requests.
See fixes of the same bug in other frameworks:
java: https://bugs.openjdk.org/browse/JDK-8283544
dart: https://github.com/dart-lang/shelf/pull/180


XenoAmess:
4.4.x backport pr added in https://github.com/apache/httpcomponents-core/pull/397

ok2c:
@XenoAmess The patch is wrong. The protocol interceptor populates the `Content-Length` or `Transfer-Encoding` _only_ if the caller has explicitly provided an entity.

PR #397, opened 2023-03-19
[HTTPCORE-736]HttpClient GET method adds Content-Length: 0 header [backport 4.4.x]

XenoAmess:
see https://github.com/apache/httpcomponents-core/pull/396


ok2c:
@XenoAmess This change is unnecessary. In HttpCore 4.x `GET`, `HEAD` and `DELETE` methods are non-entity enclosing by default.

PR #398, opened 2023-03-19
Return emptySet or emptyList instead of null to honor contract for re…

sandeepnkulkarni:
…turn type


garydgregory:
Should the collection returned be editable?

sandeepnkulkarni:
Majority of the changes proposed here are around method [getTrailerNames()](https://github.com/apache/httpcomponents-core/blob/master/httpcore5/src/main/java/org/apache/hc/core5/http/EntityDetails.java#L66) in EntityDetails interface.

There are many places where we are already returning immutable collection for implementation of this method. I believe that is what is correct. But few places were returning null. This PR fixes those issues.

ok2c:
@sandeepnkulkarni Please change the target branch to 5.3.x.

PR #399, opened 2023-03-29
HTTPCORE-737: Add maven.compiler.release if build jdk >= 9

akatkere:
This change makes sure if build jdk is based on Java >=9 API compatibility with Java 8 is maintained. Current source and target flags make sure source is Java 8 and generated byte code is Java 8 but it does not the package is compiled against Java 8 API.

Tested with both Java 8 and Java 11 build jdk. And tested Java 11 generated jar in a Java 8 runtime.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -84,6 +84,18 @@
     <hc.animal-sniffer.signature.ignores>javax.net.ssl.SSLEngine,javax.net.ssl.SSLParameters,java.nio.ByteBuffer,java.nio.CharBuffer</hc.animal-sniffer.signature.ignores>
   </properties>
 
+  <profiles>
+    <profile>
+      <id>java-8-api</id>
+      <activation>
+        <jdk>[9,)</jdk>
+      </activation>
+      <properties>
+        <maven.compiler.release>8</maven.compiler.release>
+      </properties>
+    </profile>
+  </profiles>
+
   <dependencyManagement>
     <dependencies>
       <dependency>


ok2c:
@akatkere Unfortunately this approach is not compatible with Maven Toolchain and fails when building the project with Java 11 or above and Maven Toolchain. The problem is that the activation gets triggered for the host JDK and not that used to compile the artifacts.

```
[INFO] --- maven-compiler-plugin:3.10.1:compile (default-compile) @ httpcore5 ---
[INFO] Toolchain in maven-compiler-plugin: JDK[/opt/java-1.8]
[INFO] Changes detected - recompiling the module!
[INFO] Compiling 514 source files to /home/oleg/src/apache.org/httpcomponents/httpcore/httpcore5/target/classes
[INFO] -------------------------------------------------------------
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] javac: invalid flag: --release
Usage: javac <options> <source files>
use -help for a list of possible options

[INFO] 1 error
[INFO] -------------------------------------------------------------
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for Apache HttpComponents Core Parent 5.2.2-SNAPSHOT:
[INFO] 
[INFO] Apache HttpComponents Core Parent .................. SUCCESS [  2.763 s]
[INFO] Apache HttpComponents Core HTTP/1.1 ................ FAILURE [  5.392 s]
[INFO] Apache HttpComponents Core HTTP/2 .................. SKIPPED
[INFO] Apache HttpComponents Core Reactive Extensions ..... SKIPPED
[INFO] Apache HttpComponents Core Integration Tests ....... SKIPPED
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  8.338 s
[INFO] Finished at: 2023-03-30T07:41:13+02:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.10.1:compile (default-compile) on project httpcore5: Compilation failure
[ERROR] javac: invalid flag: --release
[ERROR] Usage: javac <options> <source files>
[ERROR] use -help for a list of possible options
```

akatkere:
OK, in that case if we want httpcore5 to work on Java 8 runtime we just need to make sure we build with Java 8. In which csse no pom.xml changes are required.

ok2c:
> OK, in that case if we want httpcore5 to work on Java 8 runtime we just need to make sure we build with Java 8. In which case no pom.xml changes are required.

@akatkere Unfortunately it is so.

rschmitt:
> The problem is that the activation gets triggered for the host JDK and not that used to compile the artifacts.

Is this a documented behavior, or a bug in Maven? I can't understand how this behavior could ever be correct.

ok2c:
> Is this a documented behavior, or a bug in Maven? I can't understand how this behavior could ever be correct.

@rschmitt I am not sure. The behavior looks wrong to me but from the Maven perspective it might well be a feature. At any rate if the Maven Toolchain has been configured correctly the JRE based activation should likely never be required.

PR #408, opened 2023-06-01
[HTTPCORE-750] H2ConnPool creates too many conns under high load at initialization time

gaojieliu:
This code change checks the connection is ready to use or not before abandoning it.
When it is ready to use, it will use the conn to complete all the callbacks in the request queue, otherwise, create a new conn.

With the previous logic, under high load, H2ConnPool could create a lot of conns per endpoint and each conn will consume a lot of memory: a pair of input/output buffers.

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/reactor/AbstractIOSessionPool.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/reactor/AbstractIOSessionPool.java
@@ -183,7 +183,20 @@ public abstract class AbstractIOSessionPool<T> implements ModalCloseable {
             } else {
                 poolEntry.requestQueue.add(callback);
                 if (poolEntry.sessionFuture != null && poolEntry.sessionFuture.isDone()) {
-                    poolEntry.sessionFuture = null;
+                    // Check whether we should recreate a new conn or not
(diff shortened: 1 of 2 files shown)


gaojieliu:
Can we merge this fix into 5.2.1 release branch and cut a new release?

httpcore5/src/main/java/org/apache/hc/core5/reactor/AbstractIOSessionPool.java
```diff
@@ -183,7 +183,20 @@ private void getSessionInternal(
             } else {
                 poolEntry.requestQueue.add(callback);
                 if (poolEntry.sessionFuture != null && poolEntry.sessionFuture.isDone()) {
-                    poolEntry.sessionFuture = null;
+                    // Check whether we should recreate a new conn or not
+                    try {
+                        poolEntry.session = poolEntry.sessionFuture.get();
```
ok2c, line 188: @gaojieliu This is wrong. One _MUST NOT_ use a blocking call inside a callback that can be executed by an I/O reactor thread. 
There has to be a better way.

ok2c:
@gaojieliu We need to think of a better approach.

PR #410, opened 2023-06-20
[HTTPCORE-747]: add information about the minimum requirement for stable test execution 

Agorguy:
This PR updates the `BUILDING.txt` file to provide users with clear instructions regarding the minimum requirements for stable test execution.

The discussion is here: https://issues.apache.org/jira/browse/HTTPCORE-747

### DIFF
--- a/BUILDING.txt
+++ b/BUILDING.txt
@@ -14,9 +14,16 @@ http://maven.apache.org/run-maven/index.html
 
 (2) Executing test cases
 
-Execute the following command in order to compile and test the components
+Execute the following command in order to compile and test the components:
 
 mvn test
+
+To avoid intermittent test failures (i.e., flaky tests), it is recommended to have a machine or virtual machine with the following specifications:
+
+Minimum of 2 CPUs.
+Minimum of 2GB of RAM.
+
+By meeting these requirements, you can mitigate potential flaky tests and ensure a smooth testing experience.
 
 (3) Building packages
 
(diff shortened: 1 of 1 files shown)


BUILDING.txt
```diff
@@ -14,10 +14,17 @@ http://maven.apache.org/run-maven/index.html
...
 
 mvn test
 
+To avoid intermittent test failures (i.e., flaky tests), it is recommended to have a machine or virtual machine with the following specifications:
+
+Minimum of 2 CPUs.
+Minimum of 2GB of RAM.
+
+By meeting these requirements, you can mitigate potential flaky tests and ensure a smooth testing experience.
+
```
ok2c, line 27: @Agorguy Can we rephrase this sentence to make it sound a bit less frivolous and avoid the term "flaky"

```
These are the minimal requirements for stable test execution. In more resource constrained environments some tests may fail intermittently
```
or some such? 

Do tests _really_ need 2 CPUs?

ok2c (changes requested):

BUILDING.txt
Agorguy: I updated that sentence, your suggestion was great.

Yes, when running with less than 2CPU the chances of flakies appearing are greater.

ok2c: > Yes, when running with less than 2CPU the chances of flakies appearing are greater.

@Agorguy Bizarre. But it is good to know. Now, the real thing is to find out why.

PR #411, opened 2023-07-18
Lease connection using trylock with timeout of requesting a pool

karlchan-cn:
Hello,

I suggest that leasing a connection should be done no more than the designated request-time-out which is provided by the user.

And besides, beneath the method ```org.apache.http.nio.pool.AbstractNIOConnPool.processPendingRequest```
there are some sorta time-consuming actions like below.

image

Let's say the request-time-out of setting is **10ms** and the ```this.lock.lock()``` takes about **15ms** to acquire a permission to lease.
I recommend here we should try to lock with a timeout setting.

I hope this modification could help improve the logic and make me contribute a bit here.

Looking forward for your replay. Thanks.


ok2c:
@karlchan-cn This is a reasonable enhancement but it has already been made in the 5.x code line:

https://github.com/apache/httpcomponents-core/blob/master/httpcore5/src/main/java/org/apache/hc/core5/pool/StrictConnPool.java#L194

We are not going to make any changes to the 4.x code line beyond fixing security issues and major protocol violations. Please consider contributing your enhancements to the 5.x code line.

PR #415, opened 2023-08-17
Remove useless ParseException from signatures in EntityUtils

fdaugan:
ParseException can be removed from all signatures of this class. And since this class is final, there is no risk of overridden signature.


ok2c:
@fdaugan We cannot do that. It would break the binary compatibility with previous releases.

fdaugan:
Indeed, I have just seen that.

garydgregory:
Changing exceptions in method signatures does not break binary compatibility. It would break source compatibility though.

PR #419, opened 2023-08-25
Refactor testConcurrentOperations to use AtomicBooleans for outcome v…

arturobernalg:
Refactor `testConcurrentOperations` to use `AtomicBooleans` for outcome verification, enhancing test determinism and reducing exception unpredictability.


ok2c (approved):
@arturobernalg Interesting. Future callbacks are deterministic but the future itself is not?

arturobernalg:
> @arturobernalg Interesting. Future callbacks are deterministic but the future itself is not?

@ok2c 
With `` AtomicBoolean`   ensures we can deterministically track which callback was invoked first.  Maybe I'm wrong but  BasicFuture  operations are inherently non-deterministic

ok2c:
> Maybe I'm wrong but BasicFuture operations are inherently non-deterministic

@arturobernalg It was not my intent when I initially built `BasicFuture` but this is likely how it turned out to be. Please go ahead and merge the change-set.

PR #421, opened 2023-08-25
Support InetAddress throughout request execution

bpitman:
Note that this patch is probably incomplete - it does not break existing tests, but no tests were added nor updated to verify the new code.

We have a need to set the InetAddress for a request - the ip tells us where to connect and the hostname is needed for ssl verification.  We could have also done this with lots of separate conn pools and tls strategies, or maybe with a custom conn pool.  I preferred to have this per request, though (rather than per requester).

The current request builder supports setting HttpHost using InetAddress, but request doesn't support HttpHost so the InetAddress is lost when creating the request.  During execution, HttpHost is recreated - but with a subset of the info that was in the original HttpHost.  Additionally, the socket is created using a subset of the potential in HttpHost.  I've added InetAddress to HttpRequest and changed socket creation to use the address when it exists.

Ideally, we'd have more control of the Host header auto-entered by the request (to not include the port), but this isn't a technical requirement.

It wasn't clear how you're using branches.  Master seems to be 5.2.x, but the 5.3.x branch is missing 4 months up updates to master.  Though the version is 5.2.x, the javadoc entries I added say 5.3.

Also note that tests failed using java 11, but complete successfully with java 1.8.

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/http/HttpHost.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/http/HttpHost.java
@@ -84,7 +84,19 @@ public final class HttpHost implements NamedEndpoint, Serializable {
         Args.containsNoBlanks(hostname, "Host name");
         this.host = new Host(hostname, port);
         this.schemeName = scheme != null ? TextUtils.toLowerCase(scheme) : DEFAULT_SCHEME.id;
-        this.address = address;
+        this.address = updateAddress(address, hostname);
+    }
+
+    private InetAddress updateAddress(final InetAddress address, final String hostname) {
+        if (address == null) {
+            return address;
+        }
+        try {
+          return InetAddress.getByAddress(hostname, address.getAddress());
+        }
+        catch (final java.net.UnknownHostException e) { // should not be possible
+            return address;
+        }
     }
 
     /**
@@ -197,7 +209,9 @@ public final class HttpHost implements NamedEndpoint, Serializable {
      * @since 5.0
      */
     public HttpHost(final String scheme, final InetAddress address, final int port) {
-        this(scheme, Args.notNull(address,"Inet address"), address.getHostName(), port);
+        this.address = Args.notNull(address,"Inet address");
+        this.host = new Host(address.getHostName(), port);
+        this.schemeName = scheme != null ? TextUtils.toLowerCase(scheme) : DEFAULT_SCHEME.id;
     }
 
     /**
--- a/httpcore5/src/main/java/org/apache/hc/core5/http/HttpRequest.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/http/HttpRequest.java
@@ -27,6 +27,7 @@
 
 package org.apache.hc.core5.http;
 
+import java.net.InetAddress;
 import java.net.URI;
 import java.net.URISyntaxException;
 
@@ -93,6 +94,22 @@ public interface HttpRequest extends HttpMessage {
      * @since 5.0
      */
     void setAuthority(URIAuthority authority);
+
+    /**
+     * Returns address of this request message.
+     *
+     * @return  the address or {@code null}.
+     *
+     * @since 5.3
+     */
+    InetAddress getAddress();
+
+    /**
+     * Sets address of this request message.
+     *
+     * @since 5.3
+     */
+    void setAddress(InetAddress address);
 
     /**
      * Returns request URI of this request message. It may be an absolute or relative URI.
--- a/httpcore5/src/main/java/org/apache/hc/core5/http/impl/DefaultAddressResolver.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/http/impl/DefaultAddressResolver.java
@@ -27,6 +27,7 @@
 
 package org.apache.hc.core5.http.impl;
 
+import java.net.InetAddress;
 import java.net.InetSocketAddress;
 
 import org.apache.hc.core5.function.Resolver;
@@ -56,7 +57,10 @@ public final class DefaultAddressResolver implements Resolver<HttpHost, InetSock
                 port = 443;
             }
         }
-        return new InetSocketAddress(host.getHostName(), port);
+        final InetAddress address = host.getAddress();
+        return (address == null)
+          ? new InetSocketAddress(host.getHostName(), port)
+          : new InetSocketAddress(address, port);
     }
 
 }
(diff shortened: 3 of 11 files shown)


bpitman:
I ran locally with `-Djapicmp.skip=true` (needed with `package`, but not `test`).  Is there a mechanism for documenting known changes to the api?

ok2c:
@bpitman What is exactly the intent of this change? Why do you think you need an `InetAddress` at the request level? It is completely unclear from the description of the PR or from the source code.

Please note that one should use `HttpContext` in order to pass any additional custom details to the request execution pipeline. There generally should be no need to meddle with HTTP messages just for that.

bpitman:
See initial comment:
```
We have a need to set the InetAddress for a request - the ip tells us where to connect and the hostname is needed for ssl verification.
```

In netty 4.x, I could set a hostname verifier per request.  In httpcore5, I'd need to implement a tls strategy per requester.

bpitman:
Example data for a request:
```
{
  "state": "RUNNING",
  "status": "UP",
  "name": "upload",
  "hostname": "192.168.30.10",
  "ip": "192.168.30.10",
  "scheme": "https",
  "port": 8443,
  "sslHost": "upload-test.domain.com",
  "clientAuth": null,
  "healthcheck": "/healthcheck"
}
```

If using HttpContext to control this request execution, then I end up with a requester per request, correct?  Conn pool wouln't be able to manage max total connections across a group of nodes?

ok2c:
> We have a need to set the InetAddress for a request - the ip tells us where to connect and the hostname is needed for ssl verification.

@bpitman This makes no sense at the request level, what so ever. I have no idea about how Netty does it but in HC requests get executed over established connections and have generally no bearing on the connection management level. Please use a custom `AsyncClientConnectionOperator` if you have implement custom connection or TLS setup. One can use `HttpContext` to provide additional contextual details in special cases.

ok2c:
@bpitman In your particular case you do not even need any custom code. All you need to do is to set up the connection route correctly.
```
final HttpHost target = new HttpHost("https", InetAddress.getByName("192.168.30.10"), "upload-test.domain.com", 433);
final HttpClientContext clientContext = HttpClientContext.create();

final SimpleHttpRequest request = SimpleRequestBuilder.get()
        .setPath("/")
        .build();

System.out.println("Executing request " + request);
final Future<SimpleHttpResponse> future = client.execute(
        target,
        SimpleRequestProducer.create(request),
        SimpleResponseConsumer.create(),
        null,
        clientContext,
        new FutureCallback<SimpleHttpResponse>() {
```

bpitman:
It would have saved me a lot of time if this had worked.  Having InetAddress in HttpHost made it seem like it supported my needs right out of the box.  Look at my changes to see what's needed to preserve the InetAddress.

If the hostname is not in dns (which we don't always have control over):
```
[error] Caused by: java.net.UnknownHostException: upload-test.domain.com
```

Of when it does exist, it may resolve to a different node or resolve to a load balancer listening on a different port.

```
# resolved to load balancer listening on 443, not 8443
[error] Caused by: java.net.SocketTimeoutException: 30 SECONDS

# resolved to different host
# /etc/hosts: 192.168.30.11 my-upload-test.domain.com
[1398318490] [connState=ACTIVE, inbuf=[mode=OUTPUT pos=2 lim=2 cap=8192], outbuf=[mode=OUTPUT pos=134 lim=134 cap=8192], inputWindow=65533connState=ACTIVE, inbuf=[mode=OUTPUT pos=2 lim=2 cap=8192], outbuf=[mode=OUTPUT pos=134 lim=134 cap=8192], inputWindow=65533, incoming=[requestState=COMPLETE, responseState=COMPLETE, responseCommitted=true, keepAlive=true, done=false], outgoing=[], pipeline=0]
my-upload-test.domain.com/192.168.30.11:8443 exchange completed (connection kept alive)
```

bpitman:
To be clear, I'm using httpcore5, not httpclient5.  The client looks like non-blocking wrappers around around blocking code - though, maybe just poor examples?  We can't expect to have enough memory to hold the entire request nor response bodies in in memory.

ok2c:
> To be clear, I'm using httpcore5, not httpclient5.

@bpitman You should have started off by saying that. HttpClient has a by far more complex and sophisticated connection management code, while HttpCore has only rudimentary client support. I am quite certain that fully resolved `HttpHost` should cause no more DNS lookups with HttpClient. `HttpRequester` basically ignores  `InetAddress` in `HttpHost`. 

Instead of breaking the separation of responsibilities between logical layers one should improve (fix) `HttpAsyncRequester` in HttpCore.

bpitman:
Messing around with client examples.  The target ends up wrong if HttpHost is part of the request object, but works if passed in directly to client.execute() - as you suggested earlier.  I have not yet found anything that works using HttpHost through requester (directly using core5).

Using targets:
1) new HttpHost("https", "upload-test.domain.com", 443) # load balancer
2) new HttpHost("https", InetAddress.getByName("192.168.30.10"), "dummy.domain.com", 8443)
3) new HttpHost("https", InetAddress.getByName("192.168.30.10"), "upload-test.domain.com", 8443)
4) new HttpHost("https", InetAddress.getByName("192.168.30.10"), "my-upload-test.domain.com", 8443) // resolves to .11


AsyncClientHttpExchangeStreaming.java (uses core5 BasicRequestBuilder):
1) works correctly
2) UnkonwnHostException - incorrectly attempting resolution of dummy hostname
3) Connection refused/timeout - incorrectly resolves to load balancer and can't connect to 8443
4) incorrectly resolves to .11 and completes the request with wrong endpoint

AsyncClientHttpExchange.java (uses client5 SimpleRequestBuilder):
1) works correctly
2) UnkonwnHostException - incorrectly attempting resolution of dummy hostname
3) Connection refused/timeout - incorrectly resolves to load balancer and can't connect to 8443
4) incorrectly resolves to .11 and completes the request with wrong endpoint

Using the client.execute(target, ...):
1) works correctly
2) works correctly
3) works correctly
4) works correctly

I don't understand the rationale behind the logical layers that explains why this is the intended outcome.

bpitman:
The fixes in the pr address all of these use cases using core5 directly.  client5 is still broken - it needs similar fixes (rebuild the HttpHost using address info that had been set this the request builder).

bpitman:
Generally, yes.  This doesn't look like a typical builder pattern, though.

ok2c:
You are going about it all wrong. You are mixing up message composition and message routing responsibilities. Request builders build messages. They do not make connection setup and message routing decisions. The _same_ message can be routed via _different_ routes multiple times. If no routing information has been explicitly given the message router can take the host and the port from the message to build the default route, but that is it. The target endpoint of the route and the message authority can be entirely different. The change-set you are proposing neglects all that.

> Using the client.execute(target, ...):
> 
>     1. works correctly
> 
>     2. works correctly
> 
>     3. works correctly
> 
>     4. works correctly

Sure it does, because unlike the two other cases you actually correctly tell HttpClient to use the expected route to connect to the target endpoint 

Please do feel free to raise a JIRA for `AsyncRequester` improvements in core or provide a PR.

ok2c:
@bpitman Superseded by #422 Feel to free to review / test / comment upon

PR #422, opened 2023-08-26
Async requesters to support custom / resolved target endpoints

ok2c:


httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/bootstrap/H2MultiplexingRequester.java
```diff
@@ -119,7 +117,11 @@ public void setValidateAfterInactivity(final TimeValue timeValue) {
         connPool.setValidateAfterInactivity(timeValue);
     }
 
+    /**
```
garydgregory, line 120: Needs a description.

httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/bootstrap/H2MultiplexingRequester.java
```diff
@@ -128,18 +130,38 @@ public Cancellable execute(
...
     public Cancellable execute(
             final AsyncClientExchangeHandler exchangeHandler,
+            final HandlerFactory<AsyncPushConsumer> pushHandlerFactory,
             final Timeout timeout,
             final HttpContext context) {
-        return execute(exchangeHandler, null, timeout, context);
+        return execute(null, exchangeHandler, pushHandlerFactory, timeout, context);
+    }
+
+    /**
```
garydgregory, line 145: Needs a description.

garydgregory (changes requested):
I think new public APIs could use a Javadoc. Also, tests are missing.

bpitman:
Passing on intent from PR 421:  I understand that you want no more connection/tls configuration in HttpRequest beyond what can be expressed in URI.  In this case, why does the Builder have a setter for HttpHost?  It adds no functionality that isn't supported in other setters.  As an end user, it's only value is to add a resolved InetAddress.  If the address won't be used, then the setter only adds confusion.  I think it should be deprecated and removed.  If you're not going to do this, at least update the javadoc to explain that the address is lost in the request object; add a pointer to the way the request should be executed if the address needs to be preserved.

ok2c:
@bpitman 

The `#setHttpHost` method is merely a convenience method that copies the protocol scheme and the authority in one go from an existing `HttpHost`. I am not sure I agree it is useless or misleading. Please do feel free to propose a PR with Javadoc improvements.

```
    public AbstractRequestBuilder<T> setHttpHost(final HttpHost httpHost) {
        if (httpHost == null) {
            return this;
        }
        this.authority = new URIAuthority(httpHost);
        this.scheme = httpHost.getSchemeName();
        return this;
    }
```

arturobernalg (approved):
@ok2c  LGTM

PR #423, opened 2023-08-26
HTTPCORE-756  - Enforced the requirement that `https` and `http` URIs must not have an empty host identifier

arturobernalg:
This PR brings our `URIBuilder` in line with the standards set by RFC 9110, specifically Section 4.2.2 regarding HTTPS URIs.

Changes:

- Implemented a check to ensure `https` and `http` URIs do not have an empty host identifier.
[RFC](https://www.rfc-editor.org/rfc/rfc9110#name-https-uri-scheme) mandates such URIs be rejected as invalid.

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java
@@ -41,6 +41,7 @@ import java.util.List;
 
 import org.apache.hc.core5.http.HttpHost;
 import org.apache.hc.core5.http.NameValuePair;
+import org.apache.hc.core5.http.URIScheme;
 import org.apache.hc.core5.http.message.BasicNameValuePair;
 import org.apache.hc.core5.http.message.ParserCursor;
 import org.apache.hc.core5.util.Args;
@@ -306,6 +307,9 @@ public class URIBuilder {
      * Builds a {@link URI} instance.
      */
     public URI build() throws URISyntaxException {
+        if ((URIScheme.HTTPS.same(scheme) || URIScheme.HTTP.same(scheme))  && (TextUtils.isBlank(host))) {
+            throw new URISyntaxException(scheme, "http/https URI cannot have an empty host identifier");
+        }
         return new URI(buildString());
     }
 
--- a/httpcore5/src/test/java/org/apache/hc/core5/concurrent/TestBasicFuture.java
+++ b/httpcore5/src/test/java/org/apache/hc/core5/concurrent/TestBasicFuture.java
@@ -296,5 +296,4 @@ public class TestBasicFuture {
             }
         });
     }
-
 }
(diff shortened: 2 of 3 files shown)


ok2c:
@arturobernalg Is this PR ready for review?

arturobernalg:
> @arturobernalg Is this PR ready for review?

@ok2c yes. Or is there something I'm missing?

httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java
```diff
@@ -313,11 +314,14 @@ public URI build() throws URISyntaxException {
         return new URI(buildString());
     }
 
-    private String buildString() {
+    private String buildString() throws URISyntaxException {
         final StringBuilder sb = new StringBuilder();
         if (this.scheme != null) {
             sb.append(this.scheme).append(':');
         }
+        if (URIScheme.HTTPS.same(scheme) && (host == null || host.isEmpty())) {
```
ok2c, line 322: @arturobernalg Could you please use `TextUtils#isBlank` here for the host check? Looks good otherwise.

ok2c (approved):

httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java
arturobernalg: @ok2c you're right. Changed.

httpcore5/src/test/java/org/apache/hc/core5/concurrent/TestBasicFuture.java
```diff
@@ -260,41 +260,41 @@ public void cancelled() {
 
     @Test
     void testGetWithTimeout() {
+        final AtomicBoolean isFutureCompleted = new AtomicBoolean(false);
```
ok2c, line 263: @arturobernalg Looks good, but there appears to be unrelated change in the change-set. Please move it into a separate commit or a separate PR.

ok2c (changes requested):

httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java
```diff
@@ -309,11 +310,14 @@ public URI build() throws URISyntaxException {
         return new URI(buildString());
     }
 
-    private String buildString() {
+    private String buildString() throws URISyntaxException {
         final StringBuilder sb = new StringBuilder();
         if (this.scheme != null) {
             sb.append(this.scheme).append(':');
         }
+        if (URIScheme.HTTPS.same(scheme) && (TextUtils.isBlank(host))) {
```
reschke, line 318: Why is this specific to HTTPS?

arturobernalg: True. should be `HTTP` as well.

httpcore5/src/test/java/org/apache/hc/core5/concurrent/TestBasicFuture.java
arturobernalg: @ok2c. fair enough. Changed.

reschke:
I'm not sure that changing `toString()` is a good idea here.

ok2c:
@arturobernalg What you can do here is move the checks to the `#build` method and keep `#buildString` as is. That would also enable to remove that ugly try-catch from the `#toString`

ok2c:
@arturobernalg Please also do not forget to update the description as it will be used as a commit message

arturobernalg:
> @arturobernalg Please also do not forget to update the description as it will be used as a commit message

@ok2c 
Done

ok2c (approved):

PR #425, opened 2023-08-30
HTTPCORE-756 - Enhance HTTP header handling and deprecate legacy features

arturobernalg:
This PR introduces two significant changes to improve our alignment with modern HTTP standards:

- Trailer Headers with Chunked Transfer Encoding: We've added a check to ensure that trailer headers are only appended when using chunked transfer encoding. This change ensures that our library remains compliant with the HTTP standards regarding the use of trailer headers.

- Deprecation of ACCEPT_CHARSET: In line with the recommendations from the latest [RFC,](https://www.rfc-editor.org/rfc/rfc9110#name-changes-from-rfc-7231) we've deprecated the `ACCEPT_CHARSET` constant. This change signals to developers that its use is discouraged and helps guide them towards more modern practices.

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/http/message/MessageSupport.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/http/message/MessageSupport.java
@@ -36,6 +36,7 @@ import org.apache.hc.core5.http.EntityDetails;
 import org.apache.hc.core5.http.FormattedHeader;
 import org.apache.hc.core5.http.Header;
 import org.apache.hc.core5.http.HeaderElement;
+import org.apache.hc.core5.http.HeaderElements;
 import org.apache.hc.core5.http.HttpHeaders;
 import org.apache.hc.core5.http.HttpMessage;
 import org.apache.hc.core5.http.HttpResponse;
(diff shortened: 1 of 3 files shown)


httpcore5/src/main/java/org/apache/hc/core5/http/message/MessageSupport.java
```diff
@@ -151,6 +152,11 @@ public static void addContentEncodingHeader(final HttpMessage message, final Ent
     }
 
     public static void addTrailerHeader(final HttpMessage message, final EntityDetails entity) {
+        // Ensure using chunked transfer encoding
```
ok2c, line 155: @arturobernalg Please, do not. This is very wrong on many levels. First off `Transfer-Encoding` value is a not a single token, it is a comma delimited sequence, like `this, that, chunk. Secondly off, this is absolutely the wrong place to enforce the check. Thirdly it is not even necessary because the `*tContent` protocol interceptors already take care of.

ok2c (changes requested):

arturobernalg:
OK. I din´t see that.

PR #426, opened 2023-09-01
Refactor testGetWithTimeout for Reliability

arturobernalg:
- Use AtomicBoolean for future completion tracking.
- Simplify async behavior and improve test determinism.

### DIFF
--- a/httpcore5/src/test/java/org/apache/hc/core5/concurrent/TestBasicFuture.java
+++ b/httpcore5/src/test/java/org/apache/hc/core5/concurrent/TestBasicFuture.java
@@ -260,35 +260,36 @@ public class TestBasicFuture {
 
     @Test
     void testGetWithTimeout() {
+        final AtomicBoolean isFutureCompleted = new AtomicBoolean(false);
+
         final FutureCallback<String> callback = new FutureCallback<String>() {
             @Override
             public void completed(final String result) {
-                // Nothing to do
+                isFutureCompleted.set(true);
             }
 
             @Override
             public void failed(final Exception ex) {
-                // Nothing to do
+                // Nothing to do here for this example
             }
 
             @Override
             public void cancelled() {
-                // Nothing to do
+                // Nothing to do here for this example
             }
         };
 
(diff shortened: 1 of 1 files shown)


httpcore5/src/test/java/org/apache/hc/core5/concurrent/TestBasicFuture.java
```diff
@@ -260,35 +260,36 @@ public void cancelled() {
...
-                future.completed("test");
-            } catch (final InterruptedException e) {
-                future.failed(e);
-            }
-        }).start();
+        new Thread(() -> future.completed("test")).start();
 
+        // Poll until the future is completed or timeout
         assertTimeoutPreemptively(Duration.ofMillis(200), () -> {
+            while (!isFutureCompleted.get()) {
```
ok2c, line 288: @arturobernalg Is busy loop the only way to reliably test this condition?

arturobernalg: @ok2c 
A busy loop (or spinning loop) is not the only way to reliably test asynchronous conditions. In fact, in many cases, it's not the most efficient or recommended approach due to its CPU-intensive nature

if you believe the busy loop approach is more suitable for our testing environment, we can implement it.

ok2c (approved):

httpcore5/src/test/java/org/apache/hc/core5/concurrent/TestBasicFuture.java
thc202: Wouldn't a `CountDownLatch` work for this case?

arturobernalg: Yes, a `CountDownLatch` is indeed an alternative to achieve the same result.

thc202: It's not an alternative.

PR #428, opened 2023-09-06
Add validation for :path pseudo-header in HTTP/2 requests  RFC7540

arturobernalg:
This PR introduces validations for the :path pseudo-header field in accordance with the HTTP/2 specifications:

1. Ensures the :path pseudo-header field correctly represents the path and query parts of the target URI. Adheres to the "path-absolute" production. Optionally allows for a '?' character followed by the "query" production (as per Sections 3.3 and 3.4 of [[RFC3986](https://datatracker.ietf.org/doc/html/rfc3986)]).

2. Validates that requests in asterisk form have the value '*' for the :path pseudo-header field.

3. Enforces that the :path pseudo-header field is not left empty for "http" or "https" URIs.
4. Guarantees that "http" or "https" URIs without a path component include a '/' value for the :path pseudo-header field.
5.  Implements an exception for OPTIONS requests:
        If the request targets an "http" or "https" URI and lacks a path component, it must set the :path pseudo-header field to '*'.

### DIFF
--- a/httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/DefaultH2RequestConverter.java
+++ b/httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/DefaultH2RequestConverter.java
@@ -39,6 +39,7 @@ import org.apache.hc.core5.http.HttpRequest;
 import org.apache.hc.core5.http.HttpVersion;
 import org.apache.hc.core5.http.Method;
 import org.apache.hc.core5.http.ProtocolException;
+import org.apache.hc.core5.http.URIScheme;
 import org.apache.hc.core5.http.message.BasicHeader;
 import org.apache.hc.core5.http.message.BasicHttpRequest;
 import org.apache.hc.core5.http2.H2MessageConverter;
@@ -137,6 +138,7 @@ public final class DefaultH2RequestConverter implements H2MessageConverter<HttpR
             if (path == null) {
                 throw new ProtocolException("Mandatory request header '%s' not found", H2PseudoRequestHeaders.PATH);
             }
+            validatePathPseudoHeader(method, scheme, path);
         }
 
         final HttpRequest httpRequest = new BasicHttpRequest(method, path);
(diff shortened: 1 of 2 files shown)


httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/DefaultH2RequestConverter.java
```diff
@@ -208,4 +210,39 @@ public List<Header> convert(final HttpRequest message) throws HttpException {
...
+     * @param method The HTTP method of the request.
+     * @param scheme The scheme of the request.
+     * @param path The value of the {@code :path} pseudo-header field.
+     * @throws ProtocolException if the {@code :path} pseudo-header field is invalid.
+     */
+    private void validatePathPseudoHeader(final String method, final String scheme, final String path) throws ProtocolException {
+        if (URIScheme.HTTP.name().equalsIgnoreCase(scheme) || URIScheme.HTTPS.name().equalsIgnoreCase(scheme)) {
+            if (TextUtils.isBlank(path)) {
+                if (Method.OPTIONS.isSame(method)) {
+                    if (!"*".equals(path)) {
```
ok2c, line 234: @arturobernalg This looks wrong. Should not the `path` already be known to be blank in this branch?

ok2c (changes requested):
@arturobernalg What RFC have you been using as a reference for change?

arturobernalg:
> @arturobernalg What RFC have you been using as a reference for change?

@ok2c 
this one [RFC7540](https://datatracker.ietf.org/doc/html/rfc7540#section-8.1.2.3)

httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/DefaultH2RequestConverter.java
```diff
@@ -208,4 +210,39 @@ public List<Header> convert(final HttpRequest message) throws HttpException {
...
+     * @throws ProtocolException if any of the validations fail.
+     */
+    private void validatePathPseudoHeader(final String method, final String scheme, final String path) throws ProtocolException {
+        if (URIScheme.HTTP.name().equalsIgnoreCase(scheme) || URIScheme.HTTPS.name().equalsIgnoreCase(scheme)) {
+            if (TextUtils.isBlank(path)) {
+                throw new ProtocolException("':path' pseudo-header field must not be empty for 'http' or 'https' URIs");
+            } else {
+                if (Method.OPTIONS.isSame(method)) {
+                    if (!"*".equals(path)) {
+                        throw new ProtocolException("OPTIONS request for an 'http' or 'https' URI without a path component must have a ':path' pseudo-header field with a value of '*'");
```
ok2c, line 237: @arturobernalg This still looks wrong. Why should `OPTIONS` with the `/` path should be rejected as invalid? What you want to do it reject the path if it does not start with '/'  but make an exception for `OPTIONS` if the path is `*`.

ok2c (changes requested):

httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/DefaultH2RequestConverter.java
arturobernalg: @ok2c  please check again.

ok2c (approved):

PR #430, opened 2023-09-16
HTTPCORE-756 - Implement 'If-Range' request validation as per RFC 9110

arturobernalg:
- Ensure the presence of 'Range' header when 'If-Range' is specified.
- Enforce strong validator requirements when 'If-Range' is paired with a Date.
- Exit processing early if 'Last-Modified' header is missing, ensuring strong validation adherence.

### DIFF
--- a/httpcore5-testing/src/main/java/org/apache/hc/core5/benchmark/HttpBenchmark.java
+++ b/httpcore5-testing/src/main/java/org/apache/hc/core5/benchmark/HttpBenchmark.java
@@ -57,6 +57,7 @@ import org.apache.hc.core5.http.nio.ssl.BasicClientTlsStrategy;
 import org.apache.hc.core5.http.protocol.HttpCoreContext;
 import org.apache.hc.core5.http.protocol.HttpProcessorBuilder;
 import org.apache.hc.core5.http.protocol.RequestExpectContinue;
+import org.apache.hc.core5.http.protocol.RequestIfRange;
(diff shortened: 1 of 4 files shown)


ok2c:
@arturobernalg 
1. This change-set is definitely useful, but it kind of goes counter to the overall philosophy of the library. Generally we provide higher level APIs for consumers to use and generate low level HTTP protocol headers based on request details or execution context. Consider our statement management and authentication in client and content transport in core.  However if the library consumers manually populate protocol headers we allow them to pass any rubbish they please (with a few notable exceptions) .
2. We presently do not support content rangers at all. I am not sure we ought to mess with headers users populate manually
3. If still want to contribute this request interceptor, I suppose the client module should be a better home for it, not core.

arturobernalg:
> * This change-set is definitely useful, but it kind of goes counter to the overall philosophy of the library. Generally we provide higher level APIs for consumers to use and generate low level HTTP protocol headers based on request details or execution context. Consider our statement management and authentication in client and content transport in core.  However if the library consumers manually populate protocol headers we allow them to pass any rubbish they please (with a few notable exceptions) .
> 
> * We presently do not support content rangers at all. I am not sure we ought to mess with headers users populate manually
> 
> * If still want to contribute this request interceptor, I suppose the client module should be a better home for it, not core

@ok2c, I appreciate your feedback. I'll investigate integrating this change-set into the client module while respecting the library's philosophy.

PR #435, opened 2023-09-29
HTTPCORE-759 Content-Length missing on null entity request content

chrysophylax:
Quick fix for maximum compatibility per JIRA

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/http/protocol/RequestContent.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/http/protocol/RequestContent.java
@@ -110,6 +110,10 @@ public class RequestContent implements HttpRequestInterceptor {
                 throw new ProtocolException("Content-Length header already present");
             }
         }
+        if (entity == null) {
+            request.addHeader(HttpHeaders.CONTENT_LENGTH, "0");
+            return;
+        }
         if (entity != null) {
 
             // Check for OPTIONS request with content but no Content-Type header


httpcore5/src/main/java/org/apache/hc/core5/http/protocol/RequestContent.java
```diff
@@ -110,6 +110,10 @@ public void process(final HttpRequest request, final EntityDetails entity, final
                 throw new ProtocolException("Content-Length header already present");
             }
         }
+        if (entity == null) {
```
ok2c, line 113: @chrysophylax I think this should be done for `POST`, `PUT` and `PATCH` only and not any other standard method. That would be consistent with the behavior of HC 4.x as well.

ok2c (changes requested):

ok2c:
@chrysophylax Please also make sure the build still passes all style checks. I would also kindly ask you to add a test case to `TestStandardInterceptors` for the new behavior.

httpcore5/src/main/java/org/apache/hc/core5/http/protocol/RequestContent.java
reschke: I'd actually use a denyList with those methods that are documented not to anticipate content (which are only few).

chrysophylax: Isn't a HttpEntityEnclosingRequest inherently only for these methods which the standard defines Enclosed Content semantics?

ok2c: @chrysophylax We had to drop `HttpEntityEnclosingRequest` in 5.x because RFC 723x muddied the water by declaring it was legal for all methods to enclose an entity, well, except for `TRACE`, that is.

reschke: RFC 723x did not "muddy the water" - it just documents how HTTP works in practice.

chrysophylax: Ahhh. I follow. 

@reschke are there more methods that will not anticipate content bar `TRACE`? I could modify it either way (only `POST`, `PUT`, `PATCH`) or a denyList (`TRACE`, `??`)

ok2c: @chrysophylax Please do `POST`, `PUT`, `PATCH` by default. If desired the users can always override the default behavior through a custom request interceptor.

reschke: I believe TRACE is the only method that "MUST NOT" have content. There are some ("GET" and "DELETE"?) which "SHOULD NOT", so for these I wouldn't add "Content-Length: 0", either.

chrysophylax:
I updated the existing `testRequestContentNullEntity` test  to the new expected behaviour and added a test to make sure non-PUT/POST/PATCH don't get the `Content-Length` header added.

ok2c (approved):

reschke (commented):
I still don't get why we're making those special; I would reverse the logic.

PR #440, opened 2023-10-13
HTTPCORE-756: Support for Configurable 417 Expectation Failed Response in RequestExpectContinue

arturobernalg:
This change allows HTTP/1.1 endpoints to more accurately handle the "Expect" header field as per RFC 9110 section [10.1.1](https://www.rfc-editor.org/rfc/rfc9110#name-expect). by optionally responding with a 417 status code to indicate that the unexpected expectation cannot be met.


ok2c:
@arturobernalg This does not look right to me. This is a _client-side_ request interceptor that generates the`Expect` header, not the server-side one that processes it. The logic is misplaced. Please see `HttpServerExpectationFilter` and `AsyncServerExpectationFilter`. They already handle unexpected `Expect` header values by responding back with 417.

arturobernalg:
@ok2c, you're absolutely right. My apologies for the oversight; it's an embarrassing mistake on my part.

ok2c:
> @ok2c, you're absolutely right. My apologies for the oversight; it's an embarrassing mistake on my part.

@arturobernalg There is nothing embarrassing about it. We all err.

PR #441, opened 2023-10-15
HTTPCORE-756: RFC 9112 conformance improvements

ok2c:
* Stricter parsing of response status line. The status code must consist of 3 digits.
* `Transfer-Encoding` handling and message frame validity improvements


ok2c:
@arturobernalg Please take a look.

arturobernalg (commented):
Hi @ok2c  , the  commit refers to RFC 9112, but the ticket actually calls for changes related to RFC 9110. I missing something?"
 Should we aim to make the HTTP component fully compliant with RFC9112?

ok2c:
> Should we aim to make the HTTP component fully compliant with RFC9112?

@arturobernalg Yes, we have to. I just re-used the same JIRA ticket, that is all. I see no point in creating a separate one specifically for RFC 9112.

PR #449, opened 2023-12-19
fix: URIBuilder.getFirstQueryParam throws exception when query is emty

skssxf:
Due to description of `URIBuilder.getFirstQueryParam()` it must return `null` if named parameter not found. When query is empty `this.queryParams == null` so calling `queryParams.stream()...` results in exception. So `queryParams` replaced with `getQueryParams()` which returns empty array in such case.

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java
@@ -967,7 +967,7 @@ public class URIBuilder {
      * @since 5.2
      */
     public NameValuePair getFirstQueryParam(final String name) {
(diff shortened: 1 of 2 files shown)


httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java
```diff
@@ -967,7 +967,7 @@ public List<NameValuePair> getQueryParams() {
      * @since 5.2
      */
     public NameValuePair getFirstQueryParam(final String name) {
-        return queryParams.stream().filter(e -> name.equals(e.getName())).findFirst().orElse(null);
```
ok2c, line 970: @skssxf Could you please do a null check here instead of creating an empty list here?

ok2c (changes requested):

PR #456, opened 2024-02-06
HttpCoreContext to use instance variables for standard attributes

ok2c:
Since the inception HttpComponents used objects maps to represent context and parameters that could also be linked into a hierarchy with a parent map acting as a source of default values when not present in the local map. The use of object maps to represent a bag of all sorts of things was also a very hip (and much abused) concept at that time. Think Servlet attributes. 

A hierarchy of object maps is a very flexible and powerful structure. But it also has it cost in terms of complexity and in terms of performance. While it looked like a good idea in the beginning of HC 4.x development cycle, later it turned out to be causing more harm than good as the project progressed, mainly being unnecessarily complex, lacking proper type safety and adding too much overhead on top of that. In the HC 4.3 configuration The configuration API based on object maps got replaced with a simpler framework based on immutable plain objects.

This PR now refactors `HttpContext` and related classes to use simple attributes for standard, frequently used attributes while still providing map of objects for custom user attributes.


ok2c:
@arturobernalg Please take a look.

garydgregory:
+1 great idea.
I prefer the type safety and well defined semantics.

garydgregory (approved):

arturobernalg (approved):
@ok2c 
LGTM

PR #462, opened 2024-03-20
Update the method parameter from String to CharSequence.

wangkaish:
Update the method parameter of InetAddressUtils from String to CharSequence.

Sometimes I need to pass parts of a String(create a String slice) to this method instead of splitting the String into multiple pieces.

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/net/Host.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/net/Host.java
@@ -81,7 +81,7 @@ public final class Host implements NamedEndpoint, Serializable {
                 throw URISupport.createException(s, cursor, "Expected an IPv6 closing bracket ']'");
             }
             cursor.updatePos(cursor.getPos() + 1);
-            if (!InetAddressUtils.isIPv6Address(hostName)) {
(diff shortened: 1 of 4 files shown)


ok2c (changes requested):
@wangkaish You have not run the build locally, have you? Please make sure that you also keep the old methods in place in order to ensure binary compatibility. Please deprecate the old methods in favor of the new ones

wangkaish:
Hi, @ok2c  CI indicates that version compatibility check failed, but in fact the API is compatible, what do I do in this case?

ok2c:
@wangkaish You do something like that https://github.com/ok2c/httpcomponents-core/commit/96ca1afe17af5e7e4973b21579adf9ebd5a10a69

ok2c:
> but in fact the API is compatible
@wangkaish No, it is not. It is compile compatible but binary incompatible.

wangkaish:
@ok2c There is a method named `isIPv4MappedIPv64Address` in InetAddressUtils, is that mean `isIPv4MappedIPv6Address`?

ok2c:
@wangkaish There are style-check violations. Please address them. Please also add `@since 5.3` to a;; new methods. Otherwise looks good.

PR #463, opened 2024-03-29
Avoid unnecessary volatile reads

rhernandez35:
This commit avoids repeatedly reading volatile instance fields in ClientH2PrefaceHandler and AbstractCharDataConsumer. This is important for both performance and correctness, especially in the latter class. If a caller called `setCharset` while `getCharsetDecoder()` was being called on another thread, the second thread could see NullPointerExceptions after the `charsetDecoder` instance was unexpectedly nulled out between the null check and the return of `this.charsetDecoder` in `getCharsetDecoder`.


ok2c:
@rhernandez35 Could you please correct style check violations and ideally split the change-set into two separate commits? Looks good otherwise.

PR #467, opened 2024-06-01
HTTPCORE-766: redesign of server-side request routing

ok2c:
Redesign of server-side request routing:
* support for multiple authorities (defined as hostname / port pair)
* customizable authority resolution / authority normalization
* support for router chaining


erik-wramner:
Overall this looks fine, though it was a much larger change than I had anticipated. I only have one question. The code uses `UriPatternType.URI_PATTERN` and the other types have been deprecated. I don't quite follow the reasoning there? It won't affect our code, but I don't get why that was changed.

ok2c:
> Overall this looks fine, though it was a much larger change than I had anticipated. 

@erik-wramner True. I ended up rewriting the entire request routing layer as a result.

> I only have one question. The code uses `UriPatternType.URI_PATTERN` and the other types have been deprecated. I don't quite follow the reasoning there? It won't affect our code, but I don't get why that was changed.

That is not the case. All existing `UriPatternType` are not deprecated but their matcher implementations have been made internal. They should have never been made public in the first place.

PR #470, opened 2024-06-26
Cleanup test classes

strangelookingnerd:
This PR cleans up the test classes. The proposed changes include:

**Remove superfluous exceptions in throws clauses**
Superfluous exceptions within throws clauses have negative effects on the readability and maintainability of the code.

**Test classes and methods should have default package visibility**
It is ecommended to use the default package visibility to improve readability:
>
> Test classes, test methods, and lifecycle methods are not required to be public, but they must not be private.
It is generally recommended to omit the public modifier for test classes, test methods, and lifecycle methods unless there is a technical reason for doing so – for example, when a test class is extended by a test class in another package. Another technical reason for making classes and methods public is to simplify testing on the module path when using the Java Module System.
> — JUnit5 User Guide

**Fix assertion argument orders**
To me it seems like some assertions where not correctly upgraded when JUnit5 was introduced, where the order of arguments for some `Assertions` methods was changed.

**Remove duplicate tests**
Some test methods were implemented twice (=identical) under different names in the same class.

**Make sure tests have at least one assertion**
A test case should at least have one assertion made.

### DIFF
--- a/httpcore5/src/test/java/org/apache/hc/core5/ssl/TestSSLContextBuilder.java
+++ b/httpcore5/src/test/java/org/apache/hc/core5/ssl/TestSSLContextBuilder.java
@@ -525,11 +525,11 @@ class TestSSLContextBuilder {
         try (final SSLSocket clientSocket = (SSLSocket) clientSslContext.getSocketFactory().createSocket()) {
             clientSocket.connect(new InetSocketAddress("localhost", localPort), TIMEOUT.toMillisecondsIntBound());
             clientSocket.setSoTimeout(TIMEOUT.toMillisecondsIntBound());
-
-            clientSocket.startHandshake();
-            final InputStream inputStream = clientSocket.getInputStream();
-
-            Assertions.assertThrows(IOException.class, inputStream::read);
+            Assertions.assertThrows(IOException.class, () -> {
+                clientSocket.startHandshake();
+                final InputStream inputStream = clientSocket.getInputStream();
+                inputStream.read();
+            });
         }
     }
 
--- a/httpcore5/src/test/java/org/apache/hc/core5/ssl/TestSSLContextBuilder.java
+++ b/httpcore5/src/test/java/org/apache/hc/core5/ssl/TestSSLContextBuilder.java
@@ -525,11 +525,11 @@ class TestSSLContextBuilder {
         try (final SSLSocket clientSocket = (SSLSocket) clientSslContext.getSocketFactory().createSocket()) {
             clientSocket.connect(new InetSocketAddress("localhost", localPort), TIMEOUT.toMillisecondsIntBound());
             clientSocket.setSoTimeout(TIMEOUT.toMillisecondsIntBound());
-            Assertions.assertThrows(IOException.class, () -> {
-                clientSocket.startHandshake();
-                final InputStream inputStream = clientSocket.getInputStream();
-                inputStream.read();
-            });
+
+            clientSocket.startHandshake();
+            final InputStream inputStream = clientSocket.getInputStream();
+
+            Assertions.assertThrows(IOException.class, inputStream::read);
         }
     }
 
--- a/httpcore5/src/test/java/org/apache/hc/core5/ssl/TestSSLContextBuilder.java
+++ b/httpcore5/src/test/java/org/apache/hc/core5/ssl/TestSSLContextBuilder.java
@@ -69,7 +69,7 @@ import org.junit.jupiter.api.Test;
 /**
  * Unit tests for {@link SSLContextBuilder}.
  */
-public class TestSSLContextBuilder {
+class TestSSLContextBuilder {
 
     static final String PROVIDER_SUN_JSSE = "SunJSSE";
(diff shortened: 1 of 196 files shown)


ok2c:
Oh, that is a big one.

> Make sure tests have at least one assertion

I am aware of this recommendation but I always felt it was not justified / too extreme.

Is there a chance you could split this PR into multiple commits? Can we start with  ` Test classes and methods should have default package visibility`, `Remove duplicate tests` and `Fix assertion argument orders`?

garydgregory:
I would not change public vs package visibility, but that's just me, and the convention we've been using.

strangelookingnerd:
> > Make sure tests have at least one assertion
> 
> I am aware of this recommendation but I always felt it was not justified / too extreme.

I myself am on the edge for this. I do agree that there may be some exceptions, where the goal of a test is given by its name and it does not ultimately need an assertion. However when looking at tests like these:

```java
@Test
void testConstructor() {
   new MyObject();
}
```

I personally feel that this can be improved by something like this in order to make it clear to the reader what is intended to be tested:

```java
@Test
void testConstructor() {
   Assertions.assertDoesNotThrow(() -> new MyObject());
}
```
This is a personal preference, however.


> 
> Is there a chance you could split this PR into multiple commits? Can we start with ` Test classes and methods should have default package visibility`, `Remove duplicate tests` and `Fix assertion argument orders`?

I can defintitly try it, I will force-push the changes once I'm done and ping you 👍🏼

strangelookingnerd:
> I would not change public vs package visibility, but that's just me, and the convention we've been using.

The thing is that right now there is a mixture (at least for the methods) of `public` and `package` that would get resolved. As for the classes, I think it helps a lot in terms of readability and would make it consistent with the recommendation of the JUnit Team.

garydgregory:
"Make sure tests have at least one assertion"

That's absurd IMO.

strangelookingnerd:
> "Make sure tests have at least one assertion"
> 
> That's absurd IMO.
> 
> For example a test that fixes a bug for a thrown exception does not need to assert anything, the fact that the exception is not thrown is the "assertion".

As I said, I get your point and am myself doubting the usefulnes in certain cases. But `Assertions.assertDoesNotThrow` exists for that very reason and it _can_ help to get a clearer understanding what the test is about and what is supposed to (not) happen in it.

httpcore5/src/test/java/org/apache/hc/core5/ssl/TestSSLContextBuilder.java
```diff
@@ -525,11 +525,11 @@ void testSSLHandshakeClientUnauthenticatedError() throws Exception {
         try (final SSLSocket clientSocket = (SSLSocket) clientSslContext.getSocketFactory().createSocket()) {
             clientSocket.connect(new InetSocketAddress("localhost", localPort), TIMEOUT.toMillisecondsIntBound());
             clientSocket.setSoTimeout(TIMEOUT.toMillisecondsIntBound());
-            Assertions.assertThrows(IOException.class, () -> {
-                clientSocket.startHandshake();
-                final InputStream inputStream = clientSocket.getInputStream();
-                inputStream.read();
-            });
+
```
ok2c, line 528: @strangelookingnerd I am not sure this is correct. `#startHanshake` should be inside the asserted lambda. Is there any particular reason for changing the test case other than being able to use the method reference?

ok2c: @strangelookingnerd Otherwise looks good to me

strangelookingnerd: The lambda in `Assertions.assertThrows` should only include one call that could throw an exception. So for example:

```java
@Test
void divideByZero() {
  Assertions.assertThrows(DivideByZeroException.class, () -> {
    int a = 4 / 4;
    int b = 4 / 0;
    int c = 4 / 1;
  }
}
```

Here there would be 3 possible method calls where an exception _could_ be thrown. 
My change would make sure that the exception is thrown in `InputStream#read()` and not in `SSLSocket#startHandshake()` - because this is not what is expected.

ok2c: @strangelookingnerd Again, sounds correct in theory. But in this case, different JSSE implementations behave differently even though `#startHandshake` is expected to fail, sometimes it does not until an attempt is made to read from the underlying stream. Please keep this test as it is and will happily merge the rest.

strangelookingnerd: Corrected in [960f586](https://github.com/apache/httpcomponents-core/pull/470/commits/960f586b38fb801287f1b203b0e0472fbe36d763)

ok2c:
Thank you @strangelookingnerd for this contribution. Would you be willing to do similar changes to HttpClient?

strangelookingnerd:
> Thank you @strangelookingnerd for this contribution. Would you be willing to do similar changes to HttpClient?

Already working on it :)

PR #473, opened 2024-07-19
HTTPCORE-761: support ExtendedSocketOption in socket 

kkewwei:
Support setting TCP_KEEPIDLE, TCP_KEEPINTERVAL, TCP_KEEPCOUNT in Socket.


kkewwei:
@ok2c, please have a look when you are free.

ok2c:
@kkewwei I am fine with the proposed changes. There is a problem with the build, however.

ok2c:
@kkewwei One last thing. Could you please make symmetric changes to `SocketConfig` used by the classic transport?

kkewwei:
@ok2c I have added the parameters in `SocketConfig`, but they are not used anywhere, is that ok?

ok2c:
> @ok2c I have added the parameters in `SocketConfig`, but they are not used anywhere, is that ok?

@kkewwei Why? It is being used in `HttpRequester`, `HttpServer`  and `RequestListener`. Am I missing something?

kkewwei:
> > @ok2c I have added the parameters in `SocketConfig`, but they are not used anywhere, is that ok?
> 
> @kkewwei Why? It is being used in `HttpRequester`, `HttpServer` and `RequestListener`. Am I missing something?

@ok2c, If the three parameters are set in `Socket`, I have to use `Socket.setOption()`, which is import in java9.
image

Is there any suggestion?

ok2c:
> Is there any suggestion?

@kkewwei I suppose one has to use reflection like with `SocketOptions`.

httpcore5/src/main/java/org/apache/hc/core5/util/ReflectionUtils.java
```diff
@@ -72,4 +75,35 @@ public static int determineJRELevel() {
...
+            return null;
+        }
+    }
+
+    /**
+     * object can be ServerSocket or Socket
+     *
+     * @since 5.3
+     */
+    public static <T> void setOption(final T object, final String fieldName, final T value) {
```
kkewwei, line 96: @ok2c whether  I should use `ReflectionUtils.callSetter` directly? 

But the method doesn't throw any exception, which is not suitable for me, I want an exception to be thrown if there is a problem.

I don't know whether I should change the logic of `ReflectionUtils.callSetter`: the method doesn't swallow the exception, which will have an impact on existing usage.

kkewwei:
@ok2c please have a look when you are free.

httpcore5/src/main/java/org/apache/hc/core5/http/impl/bootstrap/HttpRequester.java
```diff
@@ -82,9 +82,14 @@
 import org.apache.hc.core5.pool.PoolEntry;
 import org.apache.hc.core5.pool.PoolStats;
 import org.apache.hc.core5.util.Args;
+import org.apache.hc.core5.util.ReflectionUtils;
 import org.apache.hc.core5.util.TimeValue;
 import org.apache.hc.core5.util.Timeout;
 
+import static org.apache.hc.core5.reactor.SingleCoreIOReactor.TCP_KEEPCOUNT;
```
ok2c, line 89: @kkewwei This should be avoided. Now classic transport services are directly coupled to a async transport class

httpcore5/src/main/java/org/apache/hc/core5/reactor/SingleCoreIOReactor.java
```diff
@@ -51,8 +53,14 @@
 import org.apache.hc.core5.util.Args;
 import org.apache.hc.core5.util.Timeout;
 
-class SingleCoreIOReactor extends AbstractSingleCoreIOReactor implements ConnectionInitiator {
+import static org.apache.hc.core5.util.ReflectionUtils.getExtendedSocketOptionOrNull;
 
+@Internal
+public class SingleCoreIOReactor extends AbstractSingleCoreIOReactor implements ConnectionInitiator {
+
+    public static final String TCP_KEEPIDLE = "TCP_KEEPIDLE";
```
ok2c, line 61: @kkewwei This is not the right place. Please move those constants to a separate file and mark it internal. The `org.apache.hc.core5.io` might be a good place.

httpcore5/src/main/java/org/apache/hc/core5/util/ReflectionUtils.java
```diff
@@ -44,12 +46,13 @@ public static void callSetter(final Object object, final String setterName, fina
         }
     }
 
-    public static <T> T callGetter(final Object object, final String getterName, final Class<T> resultType) {
+    public static <T> T callGetter(final Object object, final String getterName, final Object arg, final Class argType, final Class<T> resultType) {
```
ok2c, line 49: @kkewwei Is this method actually being used anywhere besides tests? Do we need to change it?

httpcore5/src/main/java/org/apache/hc/core5/util/ReflectionUtils.java
```diff
@@ -72,4 +75,35 @@ public static int determineJRELevel() {
...
     }
 
+    /**
+     * @since 5.3
+     */
+    public static <T> SocketOption<T> getExtendedSocketOptionOrNull(final String fieldName) {
+        try {
+            final Class<?> extendedSocketOptionsClass = Class.forName("jdk.net.ExtendedSocketOptions");
+            final Field field = extendedSocketOptionsClass.getField(fieldName);
+            return (SocketOption)field.get((Object)null);
```
ok2c, line 85: @kkewwei Please suppress `unchecked` warning

httpcore5/src/main/java/org/apache/hc/core5/util/ReflectionUtils.java
```diff
@@ -72,4 +75,35 @@ public static int determineJRELevel() {
...
+     * object can be ServerSocket or Socket
+     *
+     * @since 5.3
+     */
+    public static <T> void setOption(final T object, final String fieldName, final T value) {
+        try {
+            final Class<?> serverSocketClass = object.getClass();
+            final Method setOptionMethod = serverSocketClass.getMethod("setOption", SocketOption.class, Object.class);
+            final SocketOption<Integer> tcpKeepIdle = getExtendedSocketOptionOrNull(fieldName);
+            if (tcpKeepIdle == null) {
```
ok2c, line 101: @kkewwei The choice of the name looks bad. Can you come up with a better name here?

httpcore5/src/main/java/org/apache/hc/core5/util/ReflectionUtils.java
```diff
@@ -72,4 +75,35 @@ public static int determineJRELevel() {
...
+                throw new UnsupportedOperationException(fieldName + " is not supported in the current jdk");
+            }
+            setOptionMethod.invoke(object, tcpKeepIdle, value);
+        } catch (final Exception ignore) {
```
ok2c, line 105: @kkewwei `ignore` looks like a bad name here.

httpcore5/src/test/java/org/apache/hc/core5/reactor/TestSingleCoreIOReactor.java
```diff
@@ -0,0 +1,68 @@
+/*
+ * ====================================================================
```
ok2c, line 2: @kkewwei I am not sure what this test case actually tests. It does not seem to test much of `SingleCoreIOReactor`, just a single reflective call. Drop it.

httpcore5/src/test/java/org/apache/hc/core5/util/TestReflectionUtils.java
```diff
@@ -0,0 +1,113 @@
...
+
+    @Test
+    public void testGetExtendedSocketOptionOrNull() {
+        testGetExtendedSocketOption(TCP_KEEPIDLE);
+        testGetExtendedSocketOption(TCP_KEEPINTERVAL);
+        testGetExtendedSocketOption(TCP_KEEPCOUNT);
+    }
+
+    private void testGetExtendedSocketOption(final String option) {
+        final SocketOption socketOption = getExtendedSocketOptionOrNull(option);
```
ok2c, line 57: @kkewwei Please fix `SocketOption` generic declaration.

ok2c (changes requested):
@kkewwei There is whole bunch of small things. Please take a look at your convenience.

httpcore5/src/main/java/org/apache/hc/core5/util/ReflectionUtils.java
kkewwei: It seems unreasonable to change the publish method. otheres may be depend on it.

Should I create the similar method?

kkewwei: It indeed seem unreasonable to change the publish method. others may be depend on it.

kkewwei:
@ok2c please have an another look.

kkewwei:
@ok2c It's ok now, please have a another look.

ok2c (approved):

ok2c:
@kkewwei I tweaked your code a little and moved new methods around

PR #481, opened 2024-09-13
HTTPCORE-770: Add plusAsBlank parameter to URIBuilder for dynamic configuration

arturobernalg:
This PR introduces a new `plusAsBlank` flag to the `URIBuilder` class, allowing users to dynamically configure whether the '+' character in query parameters should be interpreted as a space (' ') or remain as a plus ('+').

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java
@@ -88,6 +88,8 @@ public class URIBuilder {
     private String fragment;
     private String encodedFragment;
 
+    private boolean plusAsBlank;
+
     /**
      * Constructs an empty instance.
      */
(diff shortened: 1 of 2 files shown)


garydgregory:
I'm not sure about creating a special case for a single character. Maybe registering a Map would be more flexible. I'm not sure what should be allowed to be mapped though.

arturobernalg:
> I'm not sure about creating a special case for a single character. Maybe registering a Map would be more flexible. I'm not sure what should be allowed to be mapped though.

HI @garydgregory 
As far as I understand, according to [RFC3986](https://datatracker.ietf.org/doc/html/rfc3986), spaces in URIs should be encoded as %20, and the + character is not treated as a space. Introducing a map would add unnecessary complexity since only + is treated this way in form encoding. Now, on second thought, it may be incorrect to allow users to change the plusAsBlank value. I'm not sure anymore.

ok2c:
> Now, on second thought, it may be incorrect to allow users to change the plusAsBlank value. I'm not sure anymore.

@arturobernalg This behavior is non-conformant but it is wide-spread and very common. This is a case where we ought to be lenient.

1. There is no need to have a special constructor for this flag. Just a setter should suffice.
2. Please add `@since 5.4`

jeremystucki:
Any idea when this will be released?

ok2c:
> Any idea when this will be released?

@jeremystucki In August for sure.

PR #484, opened 2024-09-23
Refactor to use a new class CompletingFutureContribution

garydgregory:
- Reduces boilerplate when the same pattern is used


ok2c:
@garydgregory Looks good. Please re-apply with the latest style checks.

ok2c:
@garydgregory According to the latest `japicmp` this change-set broke backward (compile) compatibility.

```
[ERROR] Failed to execute goal com.github.siom79.japicmp:japicmp-maven-plugin:0.23.0:cmp (default) on project httpcore5: There is at least one incompatibility: org.apache.hc.core5.concurrent.FutureContribution:CLASS_GENERIC_TEMPLATE_CHANGED -> [Help 1]
[ERROR] 
```

garydgregory:
Hi @ok2c 
I am ok with breaking source compatibility but obviously not binary compatibility. If you feel source compatibility is important in this case, then, yeah, some changes are needed and we should consider having the build break when that happens.

ok2c:
@garydgregory I do not think breakage of compile compatibility is warranted here. Once `hc-parent` is upgraded a newer version of `japicmp` will no longer let the build to pass in such cases.

PR #490, opened 2024-09-30
HTTPCLIENT-2233- Add Metrics Collection for IOReactor Thread Pool

arturobernalg:
This PR addresses [HTTPCLIENT-2233](https://issues.apache.org/jira/browse/HTTPCLIENT-2233) by implementing a mechanism for collecting metrics on the IOReactor thread pool in Apache HttpClient. The changes introduce an IOReactorMetricsListener interface and a LoggingIOReactorMetricsListener implementation to monitor and report key metrics for better insight into reactor performance and load.


ok2c:
@arturobernalg 
1. The original reporter of [HTTPCLIENT-2233](https://issues.apache.org/jira/browse/HTTPCLIENT-2233) never properly specified what kind of metrics they needed. Can we be sure this is what users need or we end up adding more complexity with little benefit? 
2. If you decide to proceed with this PR please add missing ` @since` tags. There is a lot of new methods
3. Please note the release series of core is 5.4 not 5.5. Please correct ` @since 5.5` tags

arturobernalg:
> @arturobernalg
> 
>     1. The original reporter of [HTTPCLIENT-2233](https://issues.apache.org/jira/browse/HTTPCLIENT-2233) never properly specified what kind of metrics they needed. Can we be sure this is what users need or we end up adding more complexity with little benefit?
> 
>     2. If you decide to proceed with this PR please add missing ` @since` tags. There is a lot of new methods
> 
>     3. Please note the release series of core is 5.4 not 5.5. Please correct ` @since 5.5` tags

@ok2c 
I understand the original request was not very specific, so I tried to include metrics that would be valuable for users monitoring I/O reactor performance. The metrics added focus on active thread count, pending connections, thread pool saturation, and queue wait time, which are key indicators for system health and performance in an async environment. I believe these metrics can provide meaningful insights for users needing more control over their thread pool's performance.

httpcore5/src/main/java/org/apache/hc/core5/reactor/SingleCoreIOReactor.java
```diff
@@ -315,6 +326,18 @@ private void validateAddress(final SocketAddress address) throws UnknownHostExce
     private void processPendingConnectionRequests() {
         IOSessionRequest sessionRequest;
         for (int i = 0; i < MAX_CHANNEL_REQUESTS && (sessionRequest = this.requestQueue.poll()) != null; i++) {
+
+            // Calculate wait time safely without keeping long-lived state
+            final long waitTimeMillis = System.currentTimeMillis() - sessionRequest.getEnqueueTime();
```
ok2c, line 331: @arturobernalg Would you agree to move metrics collection inside `if (threadPoolListener != null) {`? There is no need to collect them if there is no listener.

ok2c (changes requested):

httpcore5/src/main/java/org/apache/hc/core5/reactor/SingleCoreIOReactor.java
arturobernalg: @ok2c done

ok2c (approved):

ok2c:
@arturobernalg I added missing `@since` tags and tweaked the sequence of constructor parameters a tiny bit.

PR #496, opened 2024-10-11
Move HTTP/2 illegal message header check from message converters to protocol interceptors

ok2c:
Presently HTTP/2 specific header field validation logic is backed into message converters. It makes it impossible to customize or disable. Moreover, when a HTTP/2 promise requests when get rejected as invalid due to presence of illegal headers, it cannot be correctly mapped to a handler because there is request object to start with, and as a result the error does not get correctly propagated to the caller.

This change-set corrects the problem by moving header field validation logic from message converters to the protocol processing chain, where it belongs.

This also enables us to to allow the `host` header in H2 promise requests when connecting to Apache HTTPD.


ok2c:
@arturobernalg Please review

arturobernalg (approved):
LGTM

ok2c:
Cherry-picked to `5.3.x`

PR #498, opened 2024-10-16
Add TE header validation for HTTP/2 and HTTP/1.1 requests

arturobernalg:
This PR introduces two interceptors, H2RequestTE and RequestTE, to ensure the correct validation of the TE header in both HTTP/2 and HTTP/1.1 requests:

- H2RequestTE: Ensures that the TE header in HTTP/2 requests only contains the trailers directive, as required by the protocol. Any other transfer coding present will result in a ProtocolException.

- RequestTE: Validates the TE header in HTTP/1.1 requests by:
        Ensuring the chunked transfer coding is not explicitly listed, as it is implicitly supported.
        Verifying that the Connection: TE header is present to prevent TE from being forwarded by intermediaries.

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/http/protocol/RequestTE.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/http/protocol/RequestTE.java
@@ -1,0 +1,120 @@
+/*
+ * ====================================================================
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ * ====================================================================
+ *
+ * This software consists of voluntary contributions made by many
+ * individuals on behalf of the Apache Software Foundation.  For more
+ * information on the Apache Software Foundation, please see
+ * <http://www.apache.org/>.
+ *
+ */
+
+package org.apache.hc.core5.http.protocol;
+
+import java.io.IOException;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import org.apache.hc.core5.annotation.Contract;
+import org.apache.hc.core5.annotation.ThreadingBehavior;
+import org.apache.hc.core5.http.EntityDetails;
+import org.apache.hc.core5.http.HttpException;
+import org.apache.hc.core5.http.HttpHeaders;
+import org.apache.hc.core5.http.HttpRequest;
+import org.apache.hc.core5.http.HttpRequestInterceptor;
+import org.apache.hc.core5.http.ProtocolException;
+import org.apache.hc.core5.http.message.MessageSupport;
+import org.apache.hc.core5.util.Args;
+
+
+/**
+ * HTTP protocol interceptor responsible for validating and processing the {@link HttpHeaders#TE} header field in HTTP/1.1 requests.
+ * <p>
+ * The {@link HttpHeaders#TE} header is used to indicate transfer codings the client is willing to accept and, in some cases, whether
+ * the client is willing to accept trailer fields. This interceptor ensures that the {@link HttpHeaders#TE} header does not include
+ * the {@code chunked} transfer coding and validates the presence of the {@code Connection: TE} header.
+ * <p>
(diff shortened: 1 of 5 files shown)


ok2c:
@arturobernalg It is not our job to police the users, if they manually add headers to requests. We need to ensure strict protocol conformance of headers that HttpCore protocol interceptors generate and leave those added by the users alone. The `TE` interceptors can be useful, but we should not add them to the protocol processors by default.  These checks eat into the overall execution time and transport performance. Let the users decide whether or not they want to have them. 

What we _might_ do is to have strict versions of protocol processors _in addition_ to the default ones.

ok2c:
@arturobernalg 
1. `@since` tag should be 5.4
2. Now we have similar check done twice, once in the new interceptors and another one in `H2RequestConformance`.  I think these new checks should be moved to the existing interceptors.

arturobernalg:
> @arturobernalg
> 
>     1. `@since` tag should be 5.4
> 
>     2. Now we have similar check done twice, once in the new interceptors and another one in `H2RequestConformance`.  I think these new checks should be moved to the existing interceptors.
@ok2c 
I’ve made the changes as you suggested. I’ve removed the new TE checks for HTTP/2 since they were redundant with the existing validations in H2RequestConformance. Apologies for the oversight on my part, as I had previously validated your PR with these changes.

ok2c:
> @ok2c I’ve made the changes as you suggested. I’ve removed the new TE checks for HTTP/2 since they were redundant with the existing validations in H2RequestConformance. Apologies for the oversight on my part, as I had previously validated your PR with these changes.

@arturobernalg Do not worry. This is why do reviews and cross check one another's changes.

httpcore5/src/main/java/org/apache/hc/core5/http/protocol/RequestTE.java
```diff
@@ -0,0 +1,190 @@
...
+
+    /**
+     * Validates the presence of the {@code Connection: TE} header when the {@code TE} header is present.
+     * <p>
+     * If the {@code TE} header is used, the HTTP/1.1 protocol requires that the {@code Connection} header includes the {@code TE} directive to prevent forwarding by intermediaries.
+     *
+     * @param request the HTTP request to validate
+     * @throws HttpException if the {@code Connection: TE} header is missing
+     */
+    private void validateConnectionHeader(final HttpRequest request) throws HttpException {
```
ok2c, line 163: @arturobernalg You can really optimize this bit by making use of  `MessageSupport#parseTokens`. See how this is done in `DefaultContentLengthStrategy`

https://github.com/apache/httpcomponents-core/blob/master/httpcore5/src/main/java/org/apache/hc/core5/http/impl/DefaultContentLengthStrategy.java#L76

httpcore5/src/main/java/org/apache/hc/core5/http/protocol/RequestTE.java
```diff
@@ -0,0 +1,190 @@
...
+        final Header connectionHeader = request.getFirstHeader(HttpHeaders.CONNECTION);
```
ok2c, line 164: @arturobernalg Please also note it is legal for a message to have multiple `Connection` headers. Just looking at the first one is not good enough. I think the same applies to `TE`

arturobernalg:
@ok2c  please take another look.

httpcore5/src/main/java/org/apache/hc/core5/http/protocol/RequestTE.java
```diff
@@ -0,0 +1,184 @@
...
+     * Specifically, this method ensures that:
+     * <ul>
+     * <li>The {@link HttpHeaders#TE} header does not contain the {@code chunked} transfer coding.</li>
+     * <li>The {@code trailers} directive is allowed and treated as valid.</li>
+     * </ul>
+     *
+     * @param teValue the value of the {@code TE} header
+     * @throws HttpException if the {@code TE} header contains invalid values
+     */
+    private void validateTEField(final String teValue) throws HttpException {
```
ok2c, line 125: @arturobernalg `TE` tokens can also be transmitted with multiple headers, like `Connection`. 

```
TE: this
TE: that, trailers
```

Can you change this method to use `MessageSupport#parseTokens`?

arturobernalg: done.

ok2c: > done.

@arturobernalg Your current approach results in an unnecessary creation of an array and a copy of the header value, which is wasteful. Please use `MessageSupport#parseTokens` that avoids all that.

ok2c:
@arturobernalg Please take a look https://github.com/ok2c/httpcomponents-core/commit/24a7c00154cf5806125d83886c87967ea2d8bfc4 This is all it takes. The cost is 4 booleans and two lambdas.

arturobernalg:
> @arturobernalg Please take a look [ok2c@24a7c00](https://github.com/ok2c/httpcomponents-core/commit/24a7c00154cf5806125d83886c87967ea2d8bfc4) This is all it takes. The cost is 4 booleans and two lambdas.

@ok2c You’re absolutely right. I’ve made the necessary changes as per your suggestion. Thanks for the feedback!

ok2c (approved):

PR #499, opened 2024-10-19
Add HTTP Method QUERY support

desiderantes:
This PR introduces support for the HTTP Method `QUERY`, a safe-idempotent method that has a body. Since Node added [support for the query method](https://github.com/nodejs/node/issues/51562), there is now a need for clients to support it and behave properly (like sending content length == 0 when skipping a body).

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/annotation/Experimental.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/annotation/Experimental.java
@@ -39,5 +39,5 @@ import java.lang.annotation.Target;
 @Target({ElementType.METHOD, ElementType.TYPE, ElementType.FIELD})
 @Retention(RetentionPolicy.CLASS)
 public @interface Experimental {
-    String value() default ""; // Useful for documenting the reason for the experimental status
+    String value() default ""; // Useful for documenting the rationale for the experimental status
 }
--- a/httpcore5/src/main/java/org/apache/hc/core5/annotation/Experimental.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/annotation/Experimental.java
@@ -39,5 +39,4 @@ import java.lang.annotation.Target;
 @Target({ElementType.METHOD, ElementType.TYPE, ElementType.FIELD})
 @Retention(RetentionPolicy.CLASS)
 public @interface Experimental {
-    String value() default ""; // Useful for documenting the rationale for the experimental status
 }
--- a/httpcore5/src/main/java/org/apache/hc/core5/annotation/Experimental.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/annotation/Experimental.java
@@ -39,5 +39,5 @@ import java.lang.annotation.Target;
 @Target({ElementType.METHOD, ElementType.TYPE, ElementType.FIELD})
 @Retention(RetentionPolicy.CLASS)
 public @interface Experimental {
-    String reason() default "";
+    String value() default ""; // Useful for documenting the reason for the experimental status
 }
--- a/httpcore5/src/main/java/org/apache/hc/core5/annotation/Experimental.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/annotation/Experimental.java
@@ -36,7 +36,8 @@ import java.lang.annotation.Target;
  * The field or method to which this annotation is applied is marked as experimental.
  */
 @Documented
-@Target({ElementType.METHOD, ElementType.TYPE})
+@Target({ElementType.METHOD, ElementType.TYPE, ElementType.FIELD})
 @Retention(RetentionPolicy.CLASS)
 public @interface Experimental {
+    String reason() default "";
 }
--- a/httpcore5/src/main/java/org/apache/hc/core5/http/Method.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/http/Method.java
@@ -91,7 +91,12 @@ public enum Method {
     /**
      * The HTTP {@code PATCH} method is unsafe and non-idempotent.
      */
-    PATCH(false, false);
+    PATCH(false, false),
+
+    /**
+     * The HTTP {@code QUERY} method is safe and idempotent.
+     */
+    QUERY(true, true);
 
     private final boolean safe;
     private final boolean idempotent;
--- a/httpcore5/src/main/java/org/apache/hc/core5/http/Method.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/http/Method.java
@@ -104,7 +104,8 @@ public enum Method {
      *
      * @since 5.4
      */
-    @Experimental("QUERY method is still in DRAFT status")
+    @Experimental
+    //("QUERY method is still in DRAFT status")
     QUERY(true, true);
 
     private final boolean safe;
--- a/httpcore5/src/main/java/org/apache/hc/core5/http/Method.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/http/Method.java
@@ -95,6 +95,11 @@ public enum Method {
 
     /**
      * The HTTP {@code QUERY} method is safe and idempotent.
+     * <p>
+     * {@code QUERY} is a method defined to represent a safe, idempotent request that carries a body.
+     * This allows clients to send a body while enjoying the {@code GET}-like properties (such as easy caching,
+     * safe CORS handling [if supported] and bookmarking), as well as the {@code POST}-like properties (such as
+     * being able to send a query in a richer format, and not being limited by URI length and escaping restrictions).
      *
      * @since 5.4
      */
--- a/httpcore5/src/main/java/org/apache/hc/core5/http/Method.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/http/Method.java
@@ -29,6 +29,7 @@ package org.apache.hc.core5.http;
 
 import java.util.Locale;
 
+import org.apache.hc.core5.annotation.Experimental;
 import org.apache.hc.core5.util.Args;
 
 /**
@@ -103,6 +104,7 @@ public enum Method {
      *
      * @since 5.4
      */
+    @Experimental(reason = "QUERY method is still in DRAFT status")
     QUERY(true, true);
 
     private final boolean safe;
--- a/httpcore5/src/main/java/org/apache/hc/core5/http/Method.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/http/Method.java
@@ -95,6 +95,8 @@ public enum Method {
 
     /**
      * The HTTP {@code QUERY} method is safe and idempotent.
+     *
+     * @since 5.4
      */
     QUERY(true, true);
 
--- a/httpcore5/src/main/java/org/apache/hc/core5/http/Method.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/http/Method.java
@@ -104,7 +104,7 @@ public enum Method {
      *
      * @since 5.4
      */
-    @Experimental(reason = "QUERY method is still in DRAFT status")
+    @Experimental("QUERY method is still in DRAFT status")
     QUERY(true, true);
 
     private final boolean safe;
--- a/httpcore5/src/main/java/org/apache/hc/core5/http/io/support/ClassicRequestBuilder.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/http/io/support/ClassicRequestBuilder.java
@@ -121,6 +121,10 @@ public class ClassicRequestBuilder extends AbstractRequestBuilder<ClassicHttpReq
     }
 
     /**
+     * Initializes a new {@link ClassicRequestBuilder} instance for the {@code QUERY} method.
+     *
+     * @see Method#QUERY for more information regarding the properties of the {@code QUERY} method.
+     *
      * @since 5.4
      */
     public static ClassicRequestBuilder query() {
@@ -128,6 +132,11 @@ public class ClassicRequestBuilder extends AbstractRequestBuilder<ClassicHttpReq
     }
 
     /**
+     * Initializes a new {@link ClassicRequestBuilder} instance for the {@code QUERY} method.
+     *
+     * @param uri the request URI.
+     * @see Method#QUERY for more information regarding the properties of the {@code QUERY} method.
+     *
      * @since 5.4
      */
     public static ClassicRequestBuilder query(final URI uri) {
@@ -135,6 +144,11 @@ public class ClassicRequestBuilder extends AbstractRequestBuilder<ClassicHttpReq
     }
 
     /**
+     * Initializes a new {@link ClassicRequestBuilder} instance for the {@code QUERY} method.
+     *
+     * @param uri the request URI.
+     * @see Method#QUERY for more information regarding the properties of the {@code QUERY} method.
+     *
      * @since 5.4
      */
     public static ClassicRequestBuilder query(final String uri) {
--- a/httpcore5/src/main/java/org/apache/hc/core5/http/io/support/ClassicRequestBuilder.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/http/io/support/ClassicRequestBuilder.java
@@ -120,6 +120,18 @@ public class ClassicRequestBuilder extends AbstractRequestBuilder<ClassicHttpReq
         return new ClassicRequestBuilder(Method.HEAD, uri);
     }
 
+    public static ClassicRequestBuilder query() {
+        return new ClassicRequestBuilder(Method.QUERY);
+    }
+
+    public static ClassicRequestBuilder query(final URI uri) {
+        return new ClassicRequestBuilder(Method.QUERY, uri);
+    }
+
+    public static ClassicRequestBuilder query(final String uri) {
+        return new ClassicRequestBuilder(Method.QUERY, uri);
+    }
+
     public static ClassicRequestBuilder patch() {
         return new ClassicRequestBuilder(Method.PATCH);
     }
--- a/httpcore5/src/main/java/org/apache/hc/core5/http/io/support/ClassicRequestBuilder.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/http/io/support/ClassicRequestBuilder.java
@@ -33,6 +33,7 @@ import java.nio.charset.Charset;
 import java.util.Arrays;
 import java.util.List;
 
+import org.apache.hc.core5.annotation.Experimental;
 import org.apache.hc.core5.http.ClassicHttpRequest;
 import org.apache.hc.core5.http.ContentType;
 import org.apache.hc.core5.http.Header;
@@ -127,6 +128,7 @@ public class ClassicRequestBuilder extends AbstractRequestBuilder<ClassicHttpReq
      *
      * @since 5.4
      */
+    @Experimental
     public static ClassicRequestBuilder query() {
         return new ClassicRequestBuilder(Method.QUERY);
     }
@@ -139,6 +141,7 @@ public class ClassicRequestBuilder extends AbstractRequestBuilder<ClassicHttpReq
      *
      * @since 5.4
      */
+    @Experimental
     public static ClassicRequestBuilder query(final URI uri) {
         return new ClassicRequestBuilder(Method.QUERY, uri);
     }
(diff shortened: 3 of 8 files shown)


arturobernalg:
@desiderantes  From my understanding, the QUERY method is not defined in any official HTTP specification

garydgregory (changes requested):
Hi @desiderantes 
I'd like to see all new public and protected elements come with Javadoc (with since tags).

arturobernalg:
I have a couple of concerns regarding the addition of the QUERY method:

- Lack of Standardization: The QUERY method isn't defined in any official HTTP specification (like RFC 7231). Adding a non-standard method could lead to compatibility issues and confusion, especially since users might assume it's widely recognized. Should we really be adding this as a core method, or is there a specific context or internal protocol that necessitates this?

-     Content-Enclosing Method: The change marks QUERY as a content-enclosing method, similar to POST, PUT, and PATCH. This seems unconventional since typical query operations (like GET) don't usually involve a request body. What’s the rationale behind allowing QUERY to enclose content

desiderantes:
> I have a couple of concerns regarding the addition of the QUERY method:

Sure

>     * Lack of Standardization: The QUERY method isn't defined in any official HTTP specification (like RFC 7231). Adding a non-standard method could lead to compatibility issues and confusion, especially since users might assume it's widely recognized. Should we really be adding this as a core method, or is there a specific context or internal protocol that necessitates this?

There is a draft in progress for it: https://datatracker.ietf.org/doc/draft-ietf-httpbis-safe-method-w-body/

>     * Content-Enclosing Method: The change marks QUERY as a content-enclosing method, similar to POST, PUT, and PATCH. This seems unconventional since typical query operations (like GET) don't usually involve a request body. What’s the rationale behind allowing QUERY to enclose content

The main purpose of QUERY is to be a GET with body.

desiderantes:
> Hi @desiderantes I'd like to see all new public and protected elements come with Javadoc (with since tags).

Besides a since tag, what else should the builder methods document? All other methods don't have a javadoc.

arturobernalg:
> > I have a couple of concerns regarding the addition of the QUERY method:
> 
> Sure
> 
> > ```
> > * Lack of Standardization: The QUERY method isn't defined in any official HTTP specification (like RFC 7231). Adding a non-standard method could lead to compatibility issues and confusion, especially since users might assume it's widely recognized. Should we really be adding this as a core method, or is there a specific context or internal protocol that necessitates this?
> > ```
> 
> There is a draft in progress for it: https://datatracker.ietf.org/doc/draft-ietf-httpbis-safe-method-w-body/

A draft and not a finalized standard

> 
> > ```
> > * Content-Enclosing Method: The change marks QUERY as a content-enclosing method, similar to POST, PUT, and PATCH. This seems unconventional since typical query operations (like GET) don't usually involve a request body. What’s the rationale behind allowing QUERY to enclose content
> > ```
> 
> The main purpose of QUERY is to be a GET with body.

Since the behavior around GET with a body is often debated, there's a risk of inconsistent handling across different implementations.

garydgregory:
> > Hi @desiderantes I'd like to see all new public and protected elements come with Javadoc (with since tags).
> 
> Besides a since tag, what else should the builder methods document? All other methods don't have a javadoc.

I think it would be good to carry standard Javadoc tags like param, return, throws, whatever is appropriate for that element. For the description, I like to be able to understand why I'd use one thing versus another. In this case, a query is lile a get plus a body.

desiderantes:
> A draft and not a finalized standard

I think a major web framework adding it pushes forward the need for support by a lot, since there will be services using it now.

> Since the behavior around GET with a body is often debated, there's a risk of inconsistent handling across different implementations.

To bring a quick summary, GET is for resources, QUERY is for, well, queries. You'd GET `/entities/1` but you'd QUERY `/entities` with GraphQL, JSONPath, SQL, or whatever the server accepts as a body to perform an idempotent query.

desiderantes:
> I think it would be good to carry standard Javadoc tags like param, return, throws, whatever is appropriate for that element. For the description, I like to be able to understand why I'd use one thing versus another. In this case, a query is lile a get plus a body.

Should I add similar javadoc for all methods? I feel like it would be weird to only have a couple being documented

garydgregory:
> > I have a couple of concerns regarding the addition of the QUERY method:
> 
> Sure
> 
> > ```
> > * Lack of Standardization: The QUERY method isn't defined in any official HTTP specification (like RFC 7231). Adding a non-standard method could lead to compatibility issues and confusion, especially since users might assume it's widely recognized. Should we really be adding this as a core method, or is there a specific context or internal protocol that necessitates this?
> > ```
> 
> There is a draft in progress for it: https://datatracker.ietf.org/doc/draft-ietf-httpbis-safe-method-w-body/
> 
> > ```
> > * Content-Enclosing Method: The change marks QUERY as a content-enclosing method, similar to POST, PUT, and PATCH. This seems unconventional since typical query operations (like GET) don't usually involve a request body. What’s the rationale behind allowing QUERY to enclose content
> > ```
> 
> The main purpose of QUERY is to be a GET with body.

I wonder what the likelihood of this graduating from draft status. There must be a ton of drafts that have gone nowhere...

garydgregory:
> > I think it would be good to carry standard Javadoc tags like param, return, throws, whatever is appropriate for that element. For the description, I like to be able to understand why I'd use one thing versus another. In this case, a query is lile a get plus a body.
> 
> Should I add similar javadoc for all the other builder `$method$(...)` methods? I feel like it would be weird to only have a couple being documented

I would not ask a contributor to pick up our dropped ball. Let's see what others think about the overall goal first.

desiderantes:
> > > I have a couple of concerns regarding the addition of the QUERY method:
> > 
> > 
> > Sure
> > > ```
> > > * Lack of Standardization: The QUERY method isn't defined in any official HTTP specification (like RFC 7231). Adding a non-standard method could lead to compatibility issues and confusion, especially since users might assume it's widely recognized. Should we really be adding this as a core method, or is there a specific context or internal protocol that necessitates this?
> > > ```
> > 
> > 
> > There is a draft in progress for it: https://datatracker.ietf.org/doc/draft-ietf-httpbis-safe-method-w-body/
> > > ```
> > > * Content-Enclosing Method: The change marks QUERY as a content-enclosing method, similar to POST, PUT, and PATCH. This seems unconventional since typical query operations (like GET) don't usually involve a request body. What’s the rationale behind allowing QUERY to enclose content
> > > ```
> > 
> > 
> > The main purpose of QUERY is to be a GET with body.
> 
> I wonder what the likelihood is of this graduating from draft status. There must be a ton of drafts that have gone nowhere...

I think it's likely to graduate "soon". The [document](https://github.com/httpwg/http-extensions/commits/main/draft-ietf-httpbis-safe-method-w-body.xml) is being refined lately and [discussion](https://github.com/httpwg/http-extensions/issues?q=is%3Aissue+label%3Aquery-method+) is pretty healthy

arturobernalg:
> > > > I have a couple of concerns regarding the addition of the QUERY method:
> > > 
> > > 
> > > Sure
> > > > ```
> > > > * Lack of Standardization: The QUERY method isn't defined in any official HTTP specification (like RFC 7231). Adding a non-standard method could lead to compatibility issues and confusion, especially since users might assume it's widely recognized. Should we really be adding this as a core method, or is there a specific context or internal protocol that necessitates this?
> > > > ```
> > > 
> > > 
> > > There is a draft in progress for it: https://datatracker.ietf.org/doc/draft-ietf-httpbis-safe-method-w-body/
> > > > ```
> > > > * Content-Enclosing Method: The change marks QUERY as a content-enclosing method, similar to POST, PUT, and PATCH. This seems unconventional since typical query operations (like GET) don't usually involve a request body. What’s the rationale behind allowing QUERY to enclose content
> > > > ```
> > > 
> > > 
> > > The main purpose of QUERY is to be a GET with body.
> > 
> > 
> > I wonder what the likelihood is of this graduating from draft status. There must be a ton of drafts that have gone nowhere...
> 
> I think it's likely to graduate "soon". The [document](https://github.com/httpwg/http-extensions/commits/main/draft-ietf-httpbis-safe-method-w-body.xml) is being refined lately and [discussion](https://github.com/httpwg/http-extensions/issues?q=is%3Aissue+label%3Aquery-method+) is pretty healthy


First, about the standardization aspect: relying on a draft, no matter how 'soon' you think it will graduate, is risky. We can't build a reliable core library on something that might become a standard. Drafts come and go; many have promising discussions and then disappear. If this draft does become an RFC, great—we can add support then. Until that point, we should not prematurely introduce something that carries compatibility risk without any guarantee.

As for adding the QUERY method to support use cases like GET with a body: IMO Many server implementations  can/will handle such requests inconsistently, leading to unpredictable behavior. There is no widespread consensus among major HTTP servers to support GET with a body, and adding this as a content-enclosing method doesn't align with current expectations of how safe and idempotent methods operate. 

Regarding the Javadocs, I agree with @garydgregory . If we're adding public methods, they must be properly documented according to the Apache standards, including the @since tags, parameter descriptions, and proper guidance on usage. The fact that some existing methods lack Javadocs is not an excuse to perpetuate this issue. We should be raising the bar, not repeating past shortcomings.

Finally, it would also be better if this PR came with a proper test case.  Proper tests ensure we catch edge cases and unintended consequences before this ends up in users' hands.

desiderantes:
> > > > > I have a couple of concerns regarding the addition of the QUERY method:
> > > > 
> > > > 
> > > > Sure
> > > > > ```
> > > > > * Lack of Standardization: The QUERY method isn't defined in any official HTTP specification (like RFC 7231). Adding a non-standard method could lead to compatibility issues and confusion, especially since users might assume it's widely recognized. Should we really be adding this as a core method, or is there a specific context or internal protocol that necessitates this?
> > > > > ```
> > > > 
> > > > 
> > > > There is a draft in progress for it: https://datatracker.ietf.org/doc/draft-ietf-httpbis-safe-method-w-body/
> > > > > ```
> > > > > * Content-Enclosing Method: The change marks QUERY as a content-enclosing method, similar to POST, PUT, and PATCH. This seems unconventional since typical query operations (like GET) don't usually involve a request body. What’s the rationale behind allowing QUERY to enclose content
> > > > > ```
> > > > 
> > > > 
> > > > The main purpose of QUERY is to be a GET with body.
> > > 
> > > 
> > > I wonder what the likelihood is of this graduating from draft status. There must be a ton of drafts that have gone nowhere...
> > 
> > 
> > I think it's likely to graduate "soon". The [document](https://github.com/httpwg/http-extensions/commits/main/draft-ietf-httpbis-safe-method-w-body.xml) is being refined lately and [discussion](https://github.com/httpwg/http-extensions/issues?q=is%3Aissue+label%3Aquery-method+) is pretty healthy
> 
> First, about the standardization aspect: relying on a draft, no matter how 'soon' you think it will graduate, is risky. We can't build a reliable core library on something that might become a standard. Drafts come and go; many have promising discussions and then disappear. If this draft does become an RFC, great—we can add support then. Until that point, we should not prematurely introduce something that carries compatibility risk without any guarantee.
> 
> As for adding the QUERY method to support use cases like GET with a body: IMO Many server implementations can/will handle such requests inconsistently, leading to unpredictable behavior. There is no widespread consensus among major HTTP servers to support GET with a body, and adding this as a content-enclosing method doesn't align with current expectations of how safe and idempotent methods operate.
> 
> Regarding the Javadocs, I agree with @garydgregory . If we're adding public methods, they must be properly documented according to the Apache standards, including the @SInCE tags, parameter descriptions, and proper guidance on usage. The fact that some existing methods lack Javadocs is not an excuse to perpetuate this issue. We should be raising the bar, not repeating past shortcomings.
> 
> Finally, it would also be better if this PR came with a proper test case. Proper tests ensure we catch edge cases and unintended consequences before this ends up in users' hands.

I do not understand the argument about inconsistencies of GET with bodies, this is a different method, services would either support it or not. I think that's the main point of this new method, to have a clean way to handle GET-like requests with body, without having to guess if any part of the infrastructure assumes GET don't have bodies (like a proxy stripping them) and having surprising results and failures.

arturobernalg:
> I do not understand the argument about inconsistencies of GET with bodies, this is a different method, services would either support it or not. I think that's the main point of this new method, to have a clean way to handle GET-like requests with body, without having to guess if any part of the infrastructure assumes GET don't have bodies (like a proxy stripping them) and having surprising results and failures.

Your point is understood, but the problem remains fundamentally the same.

The fact that QUERY is technically a different method doesn't eliminate the inconsistencies; it just shifts them. Introducing QUERY as a workaround for the issues with GET bodies assumes the entire HTTP ecosystem will uniformly recognize and support it. In reality, infrastructure like proxies, caching layers, and older servers are unlikely to know what to do with QUERY requests. This could lead to dropped requests, unexpected stripping of bodies, or flat-out rejections, similar to the problems you're trying to avoid with GET and a body. Simply put, anything that doesn’t explicitly recognize QUERY may mishandle it.

desiderantes:
> > I do not understand the argument about inconsistencies of GET with bodies, this is a different method, services would either support it or not. I think that's the main point of this new method, to have a clean way to handle GET-like requests with body, without having to guess if any part of the infrastructure assumes GET don't have bodies (like a proxy stripping them) and having surprising results and failures.
> 
> Your point is understood, but the problem remains fundamentally the same.
> 
> The fact that QUERY is technically a different method doesn't eliminate the inconsistencies; it just shifts them. Introducing QUERY as a workaround for the issues with GET bodies assumes the entire HTTP ecosystem will uniformly recognize and support it. In reality, infrastructure like proxies, caching layers, and older servers are unlikely to know what to do with QUERY requests. This could lead to dropped requests, unexpected stripping of bodies, or flat-out rejections, similar to the problems you're trying to avoid with GET and a body. Simply put, anything that doesn’t explicitly recognize QUERY may mishandle it.

An explicit failure mode (not supporting unknown/custom HTTP methods) is substantially better than mishandled request bodies, so I'd say it is not a shift, but an enforcement, either the method is supported/allowed to pass (and the server ends up getting the request) or it is not, and the request fails. No middle ground, no inconsistent behaviour, other than caching proxies not caching responses (which would just be a matter of the server having a misconfigured cache).

arturobernalg:
> An explicit failure mode (not supporting unknown/custom HTTP methods) is substantially better than mishandled request bodies, so I'd say it is not a shift, but an enforcement, either the method is supported/allowed to pass (and the server ends up getting the request) or it is not, and the request fails. No middle ground, no inconsistent behaviour, other than caching proxies not caching responses (which would just be a matter of the server having a misconfigured cache).

@desiderantes I disagree.

The idea that an explicit failure mode is better doesn’t hold up here. The impact of introducing a non-standard method like QUERY isn’t simply 'support or reject.' In practice, proxies, load balancers, and other intermediaries often handle unknown methods in unpredictable ways—stripping bodies, dropping headers, or routing inconsistently. These behaviors are rarely explicit or predictable.

ok2c:
@desiderantes @garydgregory @arturobernalg We can make the new method as `@Experimental` and even `@Internal` until the standardisation is complete. I personally do not see a problem here.

garydgregory:
> > An explicit failure mode (not supporting unknown/custom HTTP methods) is substantially better than mishandled request bodies, so I'd say it is not a shift, but an enforcement, either the method is supported/allowed to pass (and the server ends up getting the request) or it is not, and the request fails. No middle ground, no inconsistent behaviour, other than caching proxies not caching responses (which would just be a matter of the server having a misconfigured cache).
> 
> @desiderantes I disagree.
> 
> The idea that an explicit failure mode is better doesn’t hold up here. The impact of introducing a non-standard method like QUERY isn’t simply 'support or reject.' In practice, proxies, load balancers, and other intermediaries often handle unknown methods in unpredictable ways—stripping bodies, dropping headers, or routing inconsistently. These behaviors are rarely explicit or predictable.

I think what I'm hearing @desiderantes say is that if my stack supports QUERY, please let me use it. It's up to me as an app author to test and document what my app does and expects.

garydgregory:
> @desiderantes @garydgregory @arturobernalg We can mark the new method as `@Experimental` and even `@Internal` until the standardisation is complete. I personally do not see a problem here.

I like the idea of annotating the code with `@Experimental("RCF...")` for all new public and protected elements.

desiderantes:
Or just not merge it now. I don't know if there's a policy regarding PR status and longevity, but I could also just move this to a draft PR and 

> @desiderantes @garydgregory @arturobernalg We can mark the new method as `@Experimental` and even `@Internal` until the standardisation is complete. I personally do not see a problem here.

This makes a lot of sense, thank you. I see that the @Experimental annotation does not include any field. Is it OK if I add an "until" or "reason" field?

arturobernalg:
> @desiderantes @garydgregory @arturobernalg We can mark the new method as `@Experimental` and even `@Internal` until the standardisation is complete. I personally do not see a problem here.

@ok2c seems like a reasonable compromise.

ok2c:
> Or just not merge it now. I don't know if there's a policy regarding PR status and longevity, but I could also just move this to a draft PR and
> 
> > @desiderantes @garydgregory @arturobernalg We can mark the new method as `@Experimental` and even `@Internal` until the standardisation is complete. I personally do not see a problem here.
> 
> This makes a lot of sense, thank you. I see that the @experimental annotation does not include any field. Is it OK if I add an "until" or "reason" field?

@desiderantes Sure. `reason` sounds more generic to me, but I leave it up to you.

garydgregory:
> > Or just not merge it now. I don't know if there's a policy regarding PR status and longevity, but I could also just move this to a draft PR and
> > > @desiderantes @garydgregory @arturobernalg We can mark the new method as `@Experimental` and even `@Internal` until the standardisation is complete. I personally do not see a problem here.
> > 
> > 
> > This makes a lot of sense, thank you. I see that the @experimental annotation does not include any field. Is it OK if I add an "until" or "reason" field?
> 
> @desiderantes Sure. `reason` sounds more generic to me, but I leave it up to you.

I would make it a plain String called "value" with a default null or empty value.

ok2c:
@desiderantes `japicmp` does not like new fields added to `@Experimenal`.  Just add a comment next to it to keep `japicmp` happy.

ok2c:
@desiderantes 
```
Failed to execute goal com.github.siom79.japicmp:japicmp-maven-plugin:0.21.2:cmp (default) on project httpcore5: There is at least one incompatibility: org.apache.hc.core5.annotation.Experimental.value():METHOD_ABSTRACT_ADDED_TO_CLASS -> [Help 1]
```
It is not worth it. Leave the poor thing be and just add a comment

ok2c (approved):

arturobernalg (approved):

ok2c:
If I hear no objections I will merge this PR.

reschke:
> The fact that QUERY is technically a different method doesn't eliminate the inconsistencies; it just shifts them. Introducing QUERY as a workaround for the issues with GET bodies assumes the entire HTTP ecosystem will uniformly recognize and support it. In reality, infrastructure like proxies, caching layers, and older servers are unlikely to know what to do with QUERY requests. This could lead to dropped requests, unexpected stripping of bodies, or flat-out rejections, similar to the problems you're trying to avoid with GET and a body. Simply put, anything that doesn’t explicitly recognize QUERY may mishandle it.

That is not true. Since WebDAV, the ecosystem has learned that OPTIONS, HEAD, and GET are special (and so say the specs). All other methods have the same characteristics with respect to how request bodies are handled. And that's good, otherwise it would be impossible to introduce new methods.

QUERY has the same characteristics as any other non-special method, be it PUT, PATCH, POST or PROPFIND, PROPPATCH, SEARCH, REPORT, etc.

The only new thing here is that it's a safe request method with request body; it's the same as PROPFIND, SEARCH, or REPORT (fwiw, these should be marked as safe and idempotent here, should I open an issue and attach a PR?)

reschke:
@garydgregory ...:

> I wonder what the likelihood is of this graduating from draft status. There must be a ton of drafts that have gone nowhere...

That is true, but none (or almost none) of these were work items of the IETF HTTP Working Group.

That said, that draft indeed has moved very slowly (even considering IETF "speed"). But the HTTP WG has now finished work (for now), and has sent the draft for review to the IESG. Once it's past that, there'll be an IETF wide Last Call, and once it has passed that, it'll be published as standards track RFC.

This should happen in late summer or autumn. You can monitor progress at https://datatracker.ietf.org/doc/draft-ietf-httpbis-safe-method-w-body/.

Hope this helps...

PR #500, opened 2024-10-20
Implement handling of invalid Date header in responses

arturobernalg:
This pull request optimizes the handling of the Date header in HTTP responses by ensuring that invalid Date headers are correctly replaced with the current time when necessary, following the guidelines in [RFC 9110, Section 6.6.1](https://www.rfc-editor.org/rfc/rfc9110.html#section-6.6.1). This behavior is optional and controlled by the replaceInvalidDate flag; by default, the behavior remains unchanged, preserving the previous logic.

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/http/protocol/ResponseDate.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/http/protocol/ResponseDate.java
@@ -36,7 +36,6 @@ import org.apache.hc.core5.http.HttpException;
 import org.apache.hc.core5.http.HttpHeaders;
 import org.apache.hc.core5.http.HttpResponse;
 import org.apache.hc.core5.http.HttpResponseInterceptor;
-import org.apache.hc.core5.http.HttpStatus;
 import org.apache.hc.core5.util.Args;
 
 /**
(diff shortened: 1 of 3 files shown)


ok2c:
@arturobernalg Looks unnecessary complex. So what if the `DATE` can be successfully parsed but it represents a point in time when Jesus roamed Galilea? What good that that make? I would either always replace `DATE` with the current date/time without even looking at at the present value.

arturobernalg:
> @arturobernalg Looks unnecessary complex. So what if the `DATE` can be successfully parsed but it represents a point in time when Jesus roamed Galilea? What good that that make? I would either always replace `DATE` with the current date/time without even looking at at the present value.

@ok2c So, basically, you are suggesting to do this:                                                                                                                       
```
public void process(final HttpResponse response, final EntityDetails entity, final HttpContext context)
        throws HttpException, IOException {
    Args.notNull(response, "HTTP response");
    final int status = response.getCode();
    if (status >= HttpStatus.SC_OK) {
        response.setHeader(HttpHeaders.DATE, HttpDateGenerator.INSTANCE.getCurrentDate());
    }
}
```

ok2c:
@arturobernalg Yes, pretty much, but with a boolean flag (`alwaysReplace or some such`) like you had before. 

By the way `status >= HttpStatus.SC_OK` check can be removed, as 1xx intermediate response messages never get passed to the protocol handler along with final responses.

arturobernalg:
> By the way `status >= HttpStatus.SC_OK` check can be removed, as 1xx intermediate response messages never get passed to the protocol handler along with final responses.

@ok2c  I’ve kept the null check for the Date header to ensure that it’s always present in the response

httpcore5/src/main/java/org/apache/hc/core5/http/protocol/ResponseDate.java
```diff
@@ -46,27 +46,49 @@
...
     }
 
     @Override
     public void process(final HttpResponse response, final EntityDetails entity, final HttpContext context)
             throws HttpException, IOException {
         Args.notNull(response, "HTTP response");
-        final int status = response.getCode();
-        if (status >= HttpStatus.SC_OK &&
-            !response.containsHeader(HttpHeaders.DATE)) {
+        final Header dateHeader = response.getFirstHeader(HttpHeaders.DATE);
```
ok2c, line 89: @arturobernalg  I suppose `replaceInvalidDate` is not the best name now. `alwaysReplace`?

This bit cab be optimized a bit. If the `alwaysReplace` flat is true there is no need to do ` response.getFirstHeader(HttpHeaders.DATE)`. Its result is basically ignored. Please re-organize the logical conditions to avoid that.

ok2c (changes requested):

httpcore5/src/main/java/org/apache/hc/core5/http/protocol/ResponseDate.java
arturobernalg: @ok2c changed.

ok2c (approved):

PR #503, opened 2024-11-01
Add NoConnectionReuseStrategy

izeye:
This PR adds `NoConnectionReuseStrategy` for v5 like [`NoConnectionReuseStrategy`](https://github.com/apache/httpcomponents-core/blob/4.4.x/httpcore/src/main/java/org/apache/http/impl/NoConnectionReuseStrategy.java) in v4.


ok2c:
@izeye Please add `@since 5.4` tag to the class.

izeye:
@ok2c Thanks for the feedback!

I added the Javadoc `@since` tag in bb8c36d as you suggested.

PR #508, opened 2024-11-27
Optimize connection close logic to resolve timeout delay issue

ooeunz:
## Description

- Introduced a configuration option useRstOnTimeout in Http1Config to allow connections to be closed with an RST (Reset) signal when a timeout occurs.
- Updated HttpRequestExecutor to respect this configuration and invoke the appropriate close mode (CloseMode.IMMEDIATE for RST or Closer.closeQuietly for FIN).
- Maintained backward compatibility by defaulting to FIN-based connection closure.

This change improves flexibility and allows faster resource cleanup in timeout scenarios, aligning with the HTTP/1.1 recommendations for abnormal connection termination.

## Problem Description

There is an issue where the connection close operation takes as long as the SocketTimeout value set in the SocketConfig after a Socket timeout occurs. This behavior arises because the NioSocketImpl performs an additional poll for the duration of the SocketTimeout during the close operation. Consequently, HTTP requests may experience delays of up to twice the configured SocketTimeout value.

## Proposed Solution

This Pull Request introduces the useRstOnTimeout configuration option to mitigate this issue by enabling connections to be closed using an RST signal instead of a FIN-based closure.
- Key benefits:
  - Avoids unnecessary delays during connection closure.
  - Ensures efficient resource cleanup.
  - Maintains backward compatibility by defaulting to FIN-based closure.

_The following image shows the test results with the socket timeout value set to 5 seconds_
img-1

img-2

## Example Workflow Before Fix

1. Socket timeout occurs:
   The socket reaches the timeout value specified in the SocketTimeout configuration.
2. Connection close:
   During the close operation, the NioSocketImpl performs an additional poll for the same timeout duration.
3. Total wait time:
   HTTP requests may end up waiting for up to twice the configured SocketTimeout value.

Resolves: [HTTPCLIENT-2324](https://issues.apache.org/jira/browse/HTTPCLIENT-2324)

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/http/config/Http1Config.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/http/config/Http1Config.java
@@ -56,10 +56,11 @@ public class Http1Config {
     private final int maxHeaderCount;
     private final int maxEmptyLineCount;
     private final int initialWindowSize;
+    private final boolean useRstOnTimeout;
 
     Http1Config(final HttpVersion version, final int bufferSize, final int chunkSizeHint,
                 final Timeout waitForContinueTimeout, final int maxLineLength, final int maxHeaderCount,
-                final int maxEmptyLineCount, final int initialWindowSize) {
+                final int maxEmptyLineCount, final int initialWindowSize, final boolean useRstOnTimeout) {
         super();
         this.version = version;
         this.bufferSize = bufferSize;
@@ -69,6 +70,7 @@ public class Http1Config {
         this.maxHeaderCount = maxHeaderCount;
         this.maxEmptyLineCount = maxEmptyLineCount;
         this.initialWindowSize = initialWindowSize;
+        this.useRstOnTimeout = useRstOnTimeout;
     }
 
     /**
@@ -109,6 +111,10 @@ public class Http1Config {
         return initialWindowSize;
     }
 
+    public boolean getUseRstOnTimeout() {
+        return useRstOnTimeout;
+    }
+
     @Override
     public String toString() {
         final StringBuilder builder = new StringBuilder();
@@ -120,6 +126,7 @@ public class Http1Config {
                 .append(", maxHeaderCount=").append(maxHeaderCount)
                 .append(", maxEmptyLineCount=").append(maxEmptyLineCount)
                 .append(", initialWindowSize=").append(initialWindowSize)
+                .append(", useRstOnTimeout=").append(useRstOnTimeout)
                 .append("]");
         return builder.toString();
     }
@@ -147,6 +154,7 @@ public class Http1Config {
     private static final int INIT_MAX_HEADER_COUNT = -1;
     private static final int INIT_MAX_LINE_LENGTH = -1;
     private static final int INIT_MAX_EMPTY_LINE_COUNT = 10;
+    private static final boolean USE_RST_ON_TIMEOUT = false;
 
     public static class Builder {
 
@@ -158,6 +166,7 @@ public class Http1Config {
         private int maxHeaderCount;
         private int maxEmptyLineCount;
         private int initialWindowSize;
+        private boolean userRstOnTimeout;
 
         Builder() {
             this.version = HttpVersion.HTTP_1_1;
@@ -168,6 +177,7 @@ public class Http1Config {
             this.maxHeaderCount = INIT_MAX_HEADER_COUNT;
             this.maxEmptyLineCount = INIT_MAX_EMPTY_LINE_COUNT;
             this.initialWindowSize = INIT_WINDOW_SIZE;
+            this.userRstOnTimeout = USE_RST_ON_TIMEOUT;
         }
 
         /**
@@ -222,6 +232,11 @@ public class Http1Config {
             return this;
         }
 
+        public Builder setUserRstOnTimeout(final boolean userRstOnTimeout) {
+            this.userRstOnTimeout = userRstOnTimeout;
+            return this;
+        }
+
         public Http1Config build() {
             return new Http1Config(
                     version,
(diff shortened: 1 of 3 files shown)


httpcore5/src/main/java/org/apache/hc/core5/http/config/Http1Config.java
```diff
@@ -109,6 +111,10 @@ public int getInitialWindowSize() {
         return initialWindowSize;
     }
 
+    public boolean getUseRstOnTimeout() {
```
ok2c, line 114: @ooeunz Please add `@since 5.4` tag

httpcore5/src/main/java/org/apache/hc/core5/http/config/Http1Config.java
```diff
@@ -222,6 +232,11 @@ public Builder setInitialWindowSize(final int initialWindowSize) {
             return this;
         }
 
+        public Builder setUserRstOnTimeout(final boolean userRstOnTimeout) {
```
ok2c, line 235: @ooeunz Please add `@since 5.4` tag

ok2c (commented):
@ooeunz

httpcore5/src/main/java/org/apache/hc/core5/http/config/Http1Config.java
ooeunz: Added. Thanks.

httpcore5/src/main/java/org/apache/hc/core5/http/config/Http1Config.java
```diff
@@ -222,6 +235,14 @@ public Builder setInitialWindowSize(final int initialWindowSize) {
             return this;
         }
 
+        /**
+         * @since 5.4
```
garydgregory, line 239: I think you need a real Javadoc description that explains what this toggle does.

httpcore5/src/main/java/org/apache/hc/core5/http/impl/io/HttpRequestExecutor.java
```diff
@@ -212,7 +213,11 @@ public ClassicHttpResponse execute(
             return response;
 
         } catch (final HttpException | IOException | RuntimeException ex) {
-            Closer.closeQuietly(conn);
+            if (http1Config.getUseRstOnTimeout()) {
```
garydgregory, line 216: Is there a use case for keeping the old behavior?

garydgregory (changes requested):
Hello @ooeunz 

See my comments.

Is it possible to write a test that shows the new code does anything?

httpcore5/src/main/java/org/apache/hc/core5/http/impl/io/HttpRequestExecutor.java
ok2c: @garydgregory Graceful shutdown of TLS connections would be one. One can drop a TLS connection without a `close-notify` handshake but it is generally considered rude.

@ooeunz Why do not we always use `IMMEDIATE` close for plain (non TLS) conections by default? Would that make the new config flag redundant?

ooeunz: To maintain backward compatibility, the existing behavior has been preserved.

However, if it is determined that there are no issues with backward compatibility, I would like to make immediate close the default behavior.

httpcore5/src/main/java/org/apache/hc/core5/http/impl/io/HttpRequestExecutor.java
```diff
@@ -212,7 +213,7 @@ public ClassicHttpResponse execute(
             return response;
 
         } catch (final HttpException | IOException | RuntimeException ex) {
-            Closer.closeQuietly(conn);
```
ok2c, line 215: @ooeunz Let's do better. Could you please do `CloseMode.IMMEDIATE` for `IOException`s and `RuntimeException`s and regular close for `SSLException`s and `HttpException`s?

ok2c (changes requested):

httpcore5/src/main/java/org/apache/hc/core5/http/impl/io/HttpRequestExecutor.java
ooeunz: @ok2c Sure, but since SSLException is a subclass of IOException, I think it should be worded like this.
```
        } catch (final IOException | RuntimeException ex) {
            Closer.close(conn, CloseMode.IMMEDIATE);
            throw ex;
        } catch (final HttpException ex) {
            Closer.closeQuietly(conn);
            throw ex;
        }
```

httpcore5/src/main/java/org/apache/hc/core5/http/impl/io/HttpRequestExecutor.java
garydgregory: TY @ok2c for the explanation. I think details like this should be in the Javadoc for the getter-setter pair.

httpcore5/src/main/java/org/apache/hc/core5/http/impl/io/HttpRequestExecutor.java
ok2c: @ooeunz How about this?
```
        } catch (final HttpException | SSLException ex) {
            Closer.closeQuietly(conn);
            throw ex;
        } catch (final IOException | RuntimeException ex) {
            Closer.close(conn, CloseMode.IMMEDIATE);
            throw ex;
        }
```

ooeunz: @ok2c Thank you, I misunderstood your intention. I’ve updated the code accordingly. :)

ok2c:
@ooeunz Almost there. Please fix the style check violations and update the title of PR and will merge the change-set.

ok2c:
@ooeunz Cherry-picked to `5.3.x`

PR #512, opened 2025-01-10
Handle overflow in Deadline calculation to ensure correct behavior for large time values.

arturobernalg:
Fixes overflow handling in Deadline calculation to ensure proper behavior when adding large time values, returning Deadline.MAX_VALUE in case of overflow.

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/util/Deadline.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/util/Deadline.java
@@ -83,9 +83,8 @@ public class Deadline {
      */
     public static Deadline calculate(final long timeMillis, final TimeValue timeValue) {
         if (TimeValue.isPositive(timeValue)) {
-            // TODO handle unlikely overflow
-            final long deadline = timeMillis + timeValue.toMilliseconds();
-            return deadline < 0 ? Deadline.MAX_VALUE : Deadline.fromUnixMilliseconds(deadline);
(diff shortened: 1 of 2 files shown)


httpcore5/src/main/java/org/apache/hc/core5/util/Deadline.java
```diff
@@ -83,8 +83,11 @@ public class Deadline {
      */
     public static Deadline calculate(final long timeMillis, final TimeValue timeValue) {
         if (TimeValue.isPositive(timeValue)) {
-            // TODO handle unlikely overflow
-            final long deadline = timeMillis + timeValue.toMilliseconds();
+            final long timeToAdd = timeValue.toMilliseconds();
```
ok2c, line 86: @arturobernalg Should not this do the same, just with a bit less code. I also think `timeToAdd > 0`  and `deadline < 0` are unnecessary.

```
        if (TimeValue.isPositive(timeValue)) {
            return Deadline.fromUnixMilliseconds(timeMillis +
                    Math.min(timeValue.toMilliseconds(), Long.MAX_VALUE - timeMillis));
        }
        return Deadline.MAX_VALUE;
```

arturobernalg: @ok2c  Updated the implementation as suggested. Thanks for the feedback!

ok2c (approved):

PR #513, opened 2025-01-12
HTTPCORE-775: The SSLIOSession::write does not handle SSLEngineResult#BUFFER_OVERFLOW

reta:
The SSLIOSession::write does not handle SSLEngineResult#BUFFER_OVERFLOW. The suggested fix it to make sure there is enough place in the output buffer to store `SSLSession#getPacketBufferSize()` bytes. Closes https://issues.apache.org/jira/browse/HTTPCORE-775

### DIFF
--- a/httpcore5-testing/src/test/java/org/apache/hc/core5/testing/SSLTestContexts.java
+++ b/httpcore5-testing/src/test/java/org/apache/hc/core5/testing/SSLTestContexts.java
@@ -40,16 +40,24 @@ import javax.net.ssl.SSLContext;
 import org.apache.hc.core5.ssl.SSLContextBuilder;
 
 public final class SSLTestContexts {
-
     public static SSLContext createServerSSLContext() {
+        return createServerSSLContext(null);
+    }
+
+    public static SSLContext createServerSSLContext(final String protocol) {
         final URL keyStoreURL = SSLTestContexts.class.getResource("/test.p12");
         final String storePassword = "nopassword";
         try {
-            return SSLContextBuilder.create()
+            SSLContextBuilder sslContextBuilder = SSLContextBuilder.create()
                     .setKeyStoreType("pkcs12")
                     .loadTrustMaterial(keyStoreURL, storePassword.toCharArray())
-                    .loadKeyMaterial(keyStoreURL, storePassword.toCharArray(), storePassword.toCharArray())
-                    .build();
+                    .loadKeyMaterial(keyStoreURL, storePassword.toCharArray(), storePassword.toCharArray());
+
+            if (protocol != null) {
+                sslContextBuilder = sslContextBuilder.setProtocol(protocol);
+            }
+
+            return sslContextBuilder.build();
         } catch (final NoSuchAlgorithmException | KeyManagementException | KeyStoreException | CertificateException |
                        UnrecoverableKeyException | IOException ex) {
             throw new IllegalStateException(ex);
@@ -57,13 +65,22 @@ public final class SSLTestContexts {
     }
 
     public static SSLContext createClientSSLContext() {
+        return createClientSSLContext(null);
+    }
+
+    public static SSLContext createClientSSLContext(final String protocol) {
         final URL keyStoreURL = SSLTestContexts.class.getResource("/test.p12");
         final String storePassword = "nopassword";
         try {
-            return SSLContextBuilder.create()
+            SSLContextBuilder sslContextBuilder = SSLContextBuilder.create()
                     .setKeyStoreType("pkcs12")
-                    .loadTrustMaterial(keyStoreURL, storePassword.toCharArray())
-                    .build();
+                    .loadTrustMaterial(keyStoreURL, storePassword.toCharArray());
+
+            if (protocol != null) {
+                sslContextBuilder = sslContextBuilder.setProtocol(protocol);
+            }
+
+            return sslContextBuilder.build();
         } catch (final NoSuchAlgorithmException | KeyManagementException | KeyStoreException | CertificateException |
                        IOException ex) {
             throw new IllegalStateException(ex);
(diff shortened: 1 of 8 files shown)


httpcore5-testing/src/test/java/org/apache/hc/core5/testing/SSLTestContexts.java
```diff
@@ -40,30 +40,47 @@
...
-            return SSLContextBuilder.create()
+            SSLContextBuilder sslContextBuilder = SSLContextBuilder.create()
                     .setKeyStoreType("pkcs12")
                     .loadTrustMaterial(keyStoreURL, storePassword.toCharArray())
-                    .loadKeyMaterial(keyStoreURL, storePassword.toCharArray(), storePassword.toCharArray())
-                    .build();
+                    .loadKeyMaterial(keyStoreURL, storePassword.toCharArray(), storePassword.toCharArray());
+
+            if (protocol != null) {
+                sslContextBuilder = sslContextBuilder.setProtocol(protocol);
```
ok2c, line 57: @reta It should be OK to pass `null`  to `#setProtocol`

httpcore5-testing/src/test/java/org/apache/hc/core5/testing/nio/H2CoreTransportTest.java
```diff
@@ -55,11 +55,15 @@ abstract class H2CoreTransportTest extends HttpCoreTransportTest {
     private final H2AsyncRequesterResource clientResource;
 
     public H2CoreTransportTest(final URIScheme scheme) {
+        this(scheme, null);
+    }
+
+    public H2CoreTransportTest(final URIScheme scheme, final String protocol) {
```
ok2c, line 61: @reta Can we give the variable a  better name to make it clearer what protocol it is? `tlsVersion` or `tlsProtocol` maybe?

ok2c (approved):
@reta Awesome! Looks really good!

reta:
> @reta Awesome! Looks really good!

Thanks a lot for the review @ok2c , review comments addressed, thank you

ok2c:
@reta Cherry-picked to `5.3.x`

patrickjamesbarry:
We believe we hit this bug, and grateful for fix. However, are their any timeouts that should of had an impact on this infinite loop?

reta:
> We believe we hit this bug, and grateful for fix. However, are their any timeouts that should of had an impact on this infinite loop?

I think the loop is infinite in theory, what is happening under the hood is just buffer adjustment: should be quick and only could fail with allocation exception. I am wondering have you seen the infinite spin in the wild? Thank you.

patrickjamesbarry:
We ran into a scenario where randomly our cpu would spike and we narrowed
it down to a hung thread from IOReactor threadpool.  We do not know what
was so special about the endpoint we were hitting or what could cause this
thread to just go wild and never return to threadpool. After updating to
this version, we have not seen it again.

We updated from
Http client 5.4.2 > 5.4.3 httpcore 5.3.3 > 5.3.4

rschmitt:
This change appears to have been incorrect. CXF was hanging due to [not signaling the availability of more data](https://github.com/apache/cxf/blob/f1b2c37bd9f3606cf7ca2d5d39cc1168e94fd9e4/rt/transports/http-hc5/src/main/java/org/apache/cxf/transport/http/asyncclient/hc5/CXFHttpAsyncRequestProducer.java#L97) when `SSLIOSession#write` signaled backpressure [here](https://github.com/apache/cxf/blob/main/rt/transports/http-hc5/src/main/java/org/apache/cxf/transport/http/asyncclient/hc5/CXFHttpAsyncRequestProducer.java#L75). The return value here (ultimately, `SSLEngineResult#bytesConsumed`) is simply being discarded.

This PR appears to have introduced dysfunctional backpressure handling into `SSLIOSession` by simply expanding the buffer as much as necessary to hold the encrypted `src`. I've seen pretty convincing evidence from another team that this was the root cause of a significant regression in heap usage and latency that cropped up in httpcore 5.3.4 and has never been fixed. Finally, when I checked out this commit and reverted the `SSLIOSession` changes, all the tests passed. I'm not sure this was ever a bug in anything except CXF.

ok2c:
@reta I think @rschmitt findings are correct and there is defect in CXF code I was not aware of.

reta:
> @reta I think @rschmitt findings are correct and there is defect in CXF code I was not aware of.

Thanks @ok2c and @rschmitt  , I am not sure it was **only** CXF related (from other people commenting, plus we have never had issues in CXF beside this specific case with TLSv1.3). 

From the implementation perspective, I think handling `SSLEngineResult.Status.BUFFER_OVERFLOW` is sound (at least it does make sense to check the result `SSLEngineResult`). If that indeed causes the issues - we could either bound the expansion or do no expand at all (revert changes on `SSLIOSession`). 

I will take a closer look shortly and get back to you folks, thanks!

ok2c:
@reta Would you consider applying the changes proposed by @rschmitt in #674 ?

It looks very plausible CXF presently does not quite fulfill the `AsyncDataProducer` contact.

```
    /**
     * Returns the number of bytes immediately available for output.
     * This method can be used as a hint to control output events
     * of the underlying I/O session.
     * <p>
     * Please note this method should return zero if the data producer
     * is unable to produce any more data, in which case
     * {@link #produce(DataStreamChannel)} method will not get triggered.
     * The producer can resume writing out data asynchronously
     * once more data becomes available or request output readiness events
     * with {@link DataStreamChannel#requestOutput()}.
     *
     * @return the number of bytes immediately available for output
     *
     * @see #produce(DataStreamChannel)
     * @see DataStreamChannel#requestOutput()
     */
    int available();
```

reta:
> @reta Would you consider applying the changes proposed by @rschmitt in #674 ?

I will definitely will into that, thanks @ok2c

PR #517, opened 2025-03-11
 RFC 3986 conformant URI encoding policy

arturobernalg:
This pull request addresses issue [HTTPCORE-778](https://issues.apache.org/jira/browse/HTTPCORE-778?page=com.atlassian.jira.plugin.system.issuetabpanels%3Aall-tabpanel), which noted that our encoding for fragment identifiers in selectors was not fully compliant with RFC 3986. The RFC defines a fragment as:

`fragment = *( pchar / "/" / "?" )`

with:

`pchar = unreserved / pct-encoded / sub-delims / ":" / "@"`

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/net/PercentCodec.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/net/PercentCodec.java
@@ -84,6 +84,16 @@ public class PercentCodec {
         URIC.or(UNRESERVED);
     }
 
+    static final BitSet FRAGMENT_SAFE = new BitSet(256);
+    static {
+        FRAGMENT_SAFE.or(UNRESERVED);
+        FRAGMENT_SAFE.or(SUB_DELIMS);
+        FRAGMENT_SAFE.set(':');
+        FRAGMENT_SAFE.set('@');
+        FRAGMENT_SAFE.set('/');
+        FRAGMENT_SAFE.set('?');
+    }
+
     static final BitSet RFC5987_UNRESERVED = new BitSet(256);
 
     static {
@@ -111,6 +121,37 @@ public class PercentCodec {
         RFC5987_UNRESERVED.set('`');
         RFC5987_UNRESERVED.set('|');
         RFC5987_UNRESERVED.set('~');
+    }
+
+    static final BitSet PCHAR = new BitSet(256);
+    static final BitSet USERINFO = new BitSet(256);
+    static final BitSet REG_NAME = new BitSet(256);
+    static final BitSet PATH_SEGMENT = new BitSet(256);
+    static final BitSet QUERY = new BitSet(256);
+    static final BitSet FRAGMENT = new BitSet(256);
+
+    static {
+        PCHAR.or(UNRESERVED);
+        PCHAR.or(SUB_DELIMS);
+        PCHAR.set(':');
+        PCHAR.set('@');
+        USERINFO.or(UNRESERVED);
+        USERINFO.or(SUB_DELIMS);
+        USERINFO.set(':');
+        REG_NAME.or(UNRESERVED);
+        REG_NAME.or(SUB_DELIMS);
+        REG_NAME.clear('!');
+        PATH_SEGMENT.or(PCHAR);
+        QUERY.or(PCHAR);
+        QUERY.set('/');
+        QUERY.set('?');
+        FRAGMENT.or(PCHAR);
+        FRAGMENT.set('/');
+        FRAGMENT.set('?');
+        // Some sub-delims remain encoded (RFC 3986 allows them unencoded, but we choose to be strict).
+        PATH_SEGMENT.clear('(');
+        PATH_SEGMENT.clear(')');
+        PATH_SEGMENT.clear('&');
     }
 
     private static final int RADIX = 16;
--- a/httpcore5/src/main/java/org/apache/hc/core5/net/PercentCodec.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/net/PercentCodec.java
@@ -84,16 +84,6 @@ public class PercentCodec {
         URIC.or(UNRESERVED);
     }
 
-    static final BitSet FRAGMENT_SAFE = new BitSet(256);
-    static {
-        FRAGMENT_SAFE.or(UNRESERVED);
-        FRAGMENT_SAFE.or(SUB_DELIMS);
-        FRAGMENT_SAFE.set(':');
-        FRAGMENT_SAFE.set('@');
-        FRAGMENT_SAFE.set('/');
-        FRAGMENT_SAFE.set('?');
-    }
-
     static final BitSet RFC5987_UNRESERVED = new BitSet(256);
 
     static {
@@ -140,7 +130,6 @@ public class PercentCodec {
         USERINFO.set(':');
         REG_NAME.or(UNRESERVED);
         REG_NAME.or(SUB_DELIMS);
-        REG_NAME.clear('!');
         PATH_SEGMENT.or(PCHAR);
         QUERY.or(PCHAR);
         QUERY.set('/');
@@ -148,10 +137,6 @@ public class PercentCodec {
         FRAGMENT.or(PCHAR);
         FRAGMENT.set('/');
         FRAGMENT.set('?');
-        // Some sub-delims remain encoded (RFC 3986 allows them unencoded, but we choose to be strict).
-        PATH_SEGMENT.clear('(');
-        PATH_SEGMENT.clear(')');
-        PATH_SEGMENT.clear('&');
     }
 
     private static final int RADIX = 16;
--- a/httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java
@@ -306,7 +306,7 @@ public class URIBuilder {
             if (i > 0 || !rootless) {
                 buf.append(PATH_SEPARATOR);
             }
-            PercentCodec.encode(buf, segment, charset);
+            PercentCodec.encode(buf, segment, charset, PercentCodec.PATH_SEGMENT, false);
             i++;
         }
     }
@@ -356,18 +356,18 @@ public class URIBuilder {
                 } else if (this.userInfo != null) {
                     final int idx = this.userInfo.indexOf(':');
                     if (idx != -1) {
-                        PercentCodec.encode(sb, this.userInfo.substring(0, idx), this.charset);
+                        PercentCodec.encode(sb, this.userInfo.substring(0, idx), this.charset, PercentCodec.USERINFO, false);
                         sb.append(':');
-                        PercentCodec.encode(sb, this.userInfo.substring(idx + 1), this.charset);
+                        PercentCodec.encode(sb, this.userInfo.substring(idx + 1), this.charset, PercentCodec.USERINFO, false);
                     } else {
-                        PercentCodec.encode(sb, this.userInfo, this.charset);
+                        PercentCodec.encode(sb, this.userInfo, this.charset, PercentCodec.USERINFO, false);
                     }
                     sb.append("@");
                 }
                 if (InetAddressUtils.isIPv6(this.host)) {
                     sb.append("[").append(this.host).append("]");
                 } else {
-                    PercentCodec.encode(sb, this.host, this.charset);
+                    PercentCodec.encode(sb, this.host, this.charset, PercentCodec.REG_NAME, false);
                 }
                 if (this.port >= 0) {
                     sb.append(":").append(this.port);
@@ -391,14 +391,14 @@ public class URIBuilder {
                 formatQuery(sb, this.queryParams, this.charset, false);
             } else if (this.query != null) {
                 sb.append("?");
-                PercentCodec.encode(sb, this.query, this.charset, PercentCodec.URIC, false);
+                PercentCodec.encode(sb, this.query, this.charset, PercentCodec.QUERY, false);
             }
         }
         if (this.encodedFragment != null) {
             sb.append("#").append(this.encodedFragment);
         } else if (this.fragment != null) {
             sb.append("#");
-            PercentCodec.encode(sb, this.fragment, this.charset, PercentCodec.URIC, false);
+            PercentCodec.encode(sb, this.fragment, this.charset, PercentCodec.FRAGMENT, false);
         }
         return sb.toString();
     }
--- a/httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java
@@ -34,6 +34,7 @@ import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.BitSet;
 import java.util.Collections;
 import java.util.LinkedList;
 import java.util.List;
@@ -87,8 +88,35 @@ public class URIBuilder {
     private Charset charset;
     private String fragment;
     private String encodedFragment;
+    private EncodingPolicy encodingPolicy = EncodingPolicy.ALL_RESERVED;
 
     private boolean plusAsBlank;
+
+    /**
+     * Defines the encoding policy for URI components in {@link URIBuilder}.
+     * This enum controls how characters are percent-encoded when constructing a URI,
+     * allowing flexibility between strict encoding and RFC 3986-compliant behavior.
+     *
+     * @since 5.4
+     */
+    public enum EncodingPolicy {
+        /**
+         * Encodes all reserved characters, allowing only unreserved characters
+         * (ALPHA, DIGIT, "-", ".", "_", "~") to remain unencoded. This is a strict
+         * policy suitable for conservative URI production where maximum encoding
+         * is desired.
+         */
+        ALL_RESERVED,
+
+        /**
+         * Follows RFC 3986 component-specific encoding rules. For example, query and
+         * fragment components allow unreserved characters, sub-delimiters ("!", "$",
+         * "&", "'", "(", ")", "*", "+", ",", ";", "="), and additional characters
+         * (":", "@", "/", "?") to remain unencoded, as defined by {@code PercentCodec.FRAGMENT}.
+         * This policy ensures compliance with RFC 3986 while maintaining interoperability.
+         */
+        RFC_3986
+    }
 
     /**
      * Constructs an empty instance.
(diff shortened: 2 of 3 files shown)


ok2c:
@arturobernalg It makes no sense to change formatting of the fragment component and ignore all other components.

arturobernalg:
> @arturobernalg It makes no sense to change formatting of the fragment component and ignore all other components.

@ok2c  please take another look.

ok2c:
@arturobernalg What about `userinfo`, `reg-name`, path `segment?`

arturobernalg:
> @arturobernalg What about `userinfo`, `reg-name`, path `segment?`

@ok2c  here's a consolidated fix ensuring consistent RFC 3986–compliant encoding for userinfo, reg-name, path segments, query, and fragment.

httpcore5/src/main/java/org/apache/hc/core5/net/PercentCodec.java
```diff
@@ -84,6 +84,16 @@ public class PercentCodec {
         URIC.or(UNRESERVED);
     }
 
+    static final BitSet FRAGMENT_SAFE = new BitSet(256);
```
ok2c, line 87: @arturobernalg `FRAGMENT_SAFE` does not seem to be used anymore. As well as  `URIC`. They should probably be removed.

ok2c (changes requested):
@arturobernalg Do you mind investing some more effort into it and adding a new attribute `encodingPolicy` as an enum of `ALL_RESERVED|RFC_3986` or similar and encoding components based on that policy?

httpcore5/src/main/java/org/apache/hc/core5/net/PercentCodec.java
arturobernalg: Removed

arturobernalg:
> > @arturobernalg What about `userinfo`, `reg-name`, path `segment?`
> 
> @ok2c here's a consolidated fix ensuring consistent RFC 3986–compliant encoding for userinfo, reg-name, path segments, query, and fragment.

@ok2c  Done. Please do another pas

httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java
```diff
@@ -356,18 +399,22 @@ private String buildString() {
                 } else if (this.userInfo != null) {
                     final int idx = this.userInfo.indexOf(':');
                     if (idx != -1) {
-                        PercentCodec.encode(sb, this.userInfo.substring(0, idx), this.charset);
+                        PercentCodec.encode(sb, this.userInfo.substring(0, idx), this.charset,
+                                encodingPolicy == EncodingPolicy.ALL_RESERVED ? PercentCodec.UNRESERVED : PercentCodec.URIC, false);
```
ok2c, line 403: @arturobernalg Should not it be `USERINFO`, not `URIC`? Same below.

httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java
```diff
@@ -391,14 +438,16 @@ private String buildString() {
                 formatQuery(sb, this.queryParams, this.charset, false);
             } else if (this.query != null) {
                 sb.append("?");
-                PercentCodec.encode(sb, this.query, this.charset, PercentCodec.URIC, false);
+                PercentCodec.encode(sb, this.query, this.charset,
+                        encodingPolicy == EncodingPolicy.ALL_RESERVED ? PercentCodec.UNRESERVED : PercentCodec.FRAGMENT, false);
```
ok2c, line 442: @arturobernalg Careful there. Here we are encoding a complete query that may use sub-delims. Even in `ALL_RESERVED` mode it should still be `URIC`.

I suppose this line should be 
```
encodingPolicy == EncodingPolicy.ALL_RESERVED ? PercentCodec.URIC : PercentCodec.QUERY, false)
```

ok2c (changes requested):
@arturobernalg Do you mind keeping `ALL_RESERVED` mode for compatibility with the behavior of previous versions?

peterhalicky:
The initial motivation for HTTPCORE-778 was the need to put some application/x-www-form-urlencoded formatted values into the fragment part of a URI. The format is pretty much the same as what is used in the query component of the URI, i.e. contains one or more key=value assignments separated by '&'.

With the above changes, if I read the diff correctly, we'll be able to use '&' as an unencoded separator. However, the keys/values (data) should be fully encoded as they can possibly contain the '&' character, and there likely isn't a way to create such URI -- either it will encode all the separators, or it won't any.

I understand that this probably opens a can of worms as it likely would require functionality similar to what is available for query parameters, but unlike query parameters in case of fragment this is only one of many ways how fragments are used.

peterhalicky:
> @peterhalicky no matter how hard we try we just cannot make everyone happy, can we?
> 
> The only way of accommodating your request would be to add support for `List<NameValuePair> fragmentParams` and make them work the same way as the query parameters.

Yes, I agree, I'd be quite hesitant to implement this myself. For my use case I can get away with restricting the keys and values to not containing `&`. I just wanted to point out that this is still a restriction that will not allow users to create all possible URIs that would otherwise be legal as per the RFC (not sure if the library aims for that or not).

ok2c:
> I just wanted to point out that this is still a restriction that will not allow users to create all possible URIs that would otherwise be legal as per the RFC 

@peterhalicky The whole point of representing the query component as `List<NameValuePair>` is make it possible for name/value pairs to contain arbitrary characters. The same could be done for the fragment component.

peterhalicky:
> The same could be done for the fragment component.
The fragment component often contains all kinds of content. Sometimes it is `application/x-www-form-urlencoded`, other times it looks like a path, i.e. `/` separated path segments. Representing the fragment component as `List<NameValuePair>` will only support the former, but not the latter.

A possible solution could be to let the client build the fragment by parts, specifying how much encoding should be done for each part. Something like `.appendFragment("&", URIC)` and `.appendFragment("valuewith&", RESERVED)`.

httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java
```diff
@@ -87,9 +87,36 @@ public static URIBuilder loopbackAddress() {
     private Charset charset;
     private String fragment;
     private String encodedFragment;
+    private EncodingPolicy encodingPolicy = EncodingPolicy.RFC_3986;
```
ok2c, line 90: @arturobernalg Should not it be `EncodingPolicy.ALL_RESERVED` by default?

httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java
```diff
@@ -391,14 +438,16 @@ private String buildString() {
...
+                        encodingPolicy == EncodingPolicy.ALL_RESERVED ? PercentCodec.URIC : PercentCodec.QUERY, false);
             }
         }
         if (this.encodedFragment != null) {
             sb.append("#").append(this.encodedFragment);
         } else if (this.fragment != null) {
             sb.append("#");
-            PercentCodec.encode(sb, this.fragment, this.charset, PercentCodec.URIC, false);
+            PercentCodec.encode(sb, this.fragment, this.charset,
+                    encodingPolicy == EncodingPolicy.ALL_RESERVED ? PercentCodec.UNRESERVED : PercentCodec.FRAGMENT, false);
```
ok2c, line 450: @arturobernalg We should use `URIC` here for consistency with the latest changes in 5.3.x

httpcore5/src/main/java/org/apache/hc/core5/net/PercentCodec.java
```diff
@@ -113,6 +113,37 @@ public class PercentCodec {
...
+        PCHAR.or(UNRESERVED);
+        PCHAR.or(SUB_DELIMS);
+        PCHAR.set(':');
+        PCHAR.set('@');
+        USERINFO.or(UNRESERVED);
+        USERINFO.or(SUB_DELIMS);
+        USERINFO.set(':');
+        REG_NAME.or(UNRESERVED);
+        REG_NAME.or(SUB_DELIMS);
+        REG_NAME.clear('!');
```
ok2c, line 133: @arturobernalg Why clearing '!'? The reg-name is defimed as `*( unreserved / pct-encoded / sub-delims )`. The '!' should be permitted or am I missing something?

httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java
arturobernalg: done

httpcore5/src/main/java/org/apache/hc/core5/net/PercentCodec.java
```diff
@@ -113,6 +113,37 @@ public class PercentCodec {
...
+        QUERY.set('/');
+        QUERY.set('?');
+        FRAGMENT.or(PCHAR);
+        FRAGMENT.set('/');
+        FRAGMENT.set('?');
+        // Some sub-delims remain encoded (RFC 3986 allows them unencoded, but we choose to be strict).
+        PATH_SEGMENT.clear('(');
+        PATH_SEGMENT.clear(')');
+        PATH_SEGMENT.clear('&');
+    }
```
ok2c, line 145: @arturobernalg  I cannot find anything in the RFC stating that. What am I missing? What a super strict mode we have `ALL_RESERVED`

arturobernalg: > @arturobernalg I cannot find anything in the RFC stating that. What am I missing? What a super strict mode we have `ALL_RESERVED`

@olegk  The test expects `(`, `)`, `&` to remain percent-encoded in path segments (e.g., `blah-%28%20-blah-%20%26%20-blah-%20%29-blah/`), but without the `PATH_SEGMENT.clear('(')`, `clear(')')`, `clear('&')` in `PercentCodec.java`, it outputs `blah-(%20-blah-%20&%20-blah-%20)-blah/`. Reintroducing the `clear()` calls fixes it, but I’m wondering if this aligns with our intended behavior or if we should adjust `formatPath` to handle encoding differently based on `encodingPolicy`.
I noticed the older `formatPath` used `PercentCodec.encode(buf, segment, charset, false)` with `UNRESERVED`, which worked, but the new version with `encodingPolicy` support might need tweaking. Could you take a look and advise?

ok2c: @arturobernalg You have overlooked static `formatQuery` and `formatPath` methods. They should also take the encoding policy into account. Apply this patch and everything should work

```
===================================================================
diff --git a/httpcore5/src/main/java/org/apache/hc/core5/net/PercentCodec.java b/httpcore5/src/main/java/org/apache/hc/core5/net/PercentCodec.java
--- a/httpcore5/src/main/java/org/apache/hc/core5/net/PercentCodec.java	(revision 53477ecc93d331f8348a6748a918d86b25666471)
+++ b/httpcore5/src/main/java/org/apache/hc/core5/net/PercentCodec.java	(date 1742025384945)
@@ -137,10 +137,6 @@
         FRAGMENT.or(PCHAR);
         FRAGMENT.set('/');
         FRAGMENT.set('?');
-        // Some sub-delims remain encoded (RFC 3986 allows them unencoded, but we choose to be strict).
-        PATH_SEGMENT.clear('(');
-        PATH_SEGMENT.clear(')');
-        PATH_SEGMENT.clear('&');
     }
 
     private static final int RADIX = 16;
Index: httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java b/httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java
--- a/httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java	(revision 53477ecc93d331f8348a6748a918d86b25666471)
+++ b/httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java	(date 1742026397263)
@@ -34,6 +34,7 @@
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.BitSet;
 import java.util.Collections;
 import java.util.LinkedList;
 import java.util.List;
@@ -343,19 +344,25 @@
         return list;
     }
 
-    static void formatPath(final StringBuilder buf, final Iterable<String> segments, final boolean rootless, final Charset charset) {
+    static void formatPath(final StringBuilder buf, final Iterable<String> segments, final boolean rootless,
+                           final Charset charset, final BitSet safechars) {
         int i = 0;
         for (final String segment : segments) {
             if (i > 0 || !rootless) {
                 buf.append(PATH_SEPARATOR);
             }
-            PercentCodec.encode(buf, segment, charset, PercentCodec.PATH_SEGMENT, false);
+            PercentCodec.encode(buf, segment, charset, safechars, false);
             i++;
         }
     }
 
-    static void formatQuery(final StringBuilder buf, final Iterable<? extends NameValuePair> params, final Charset charset,
-                            final boolean blankAsPlus) {
+    static void formatPath(final StringBuilder buf, final Iterable<String> segments, final boolean rootless,
+                           final Charset charset) {
+        formatPath(buf, segments, rootless, charset, PercentCodec.UNRESERVED);
+    }
+
+    static void formatQuery(final StringBuilder buf, final Iterable<? extends NameValuePair> params,
+                            final Charset charset, final BitSet safechars, final boolean blankAsPlus) {
         int i = 0;
         for (final NameValuePair parameter : params) {
             if (i > 0) {
@@ -364,12 +371,17 @@
             PercentCodec.encode(buf, parameter.getName(), charset, blankAsPlus);
             if (parameter.getValue() != null) {
                 buf.append(PARAM_VALUE_SEPARATOR);
-                PercentCodec.encode(buf, parameter.getValue(), charset, blankAsPlus);
+                PercentCodec.encode(buf, parameter.getValue(), charset, safechars, blankAsPlus);
             }
             i++;
         }
     }
 
+    static void formatQuery(final StringBuilder buf, final Iterable<? extends NameValuePair> params,
+                            final Charset charset, final boolean blankAsPlus) {
+        formatQuery(buf, params, charset, PercentCodec.UNRESERVED, blankAsPlus);
+    }
+
     /**
      * Builds a {@link URI} instance.
      */
@@ -429,13 +441,15 @@
                 }
                 sb.append(this.encodedPath);
             } else if (this.pathSegments != null) {
-                formatPath(sb, this.pathSegments, !authoritySpecified && this.pathRootless, this.charset);
+                formatPath(sb, this.pathSegments, !authoritySpecified && this.pathRootless, this.charset,
+                        encodingPolicy == EncodingPolicy.ALL_RESERVED ? PercentCodec.UNRESERVED : PercentCodec.PATH_SEGMENT);
             }
             if (this.encodedQuery != null) {
                 sb.append("?").append(this.encodedQuery);
             } else if (this.queryParams != null && !this.queryParams.isEmpty()) {
                 sb.append("?");
-                formatQuery(sb, this.queryParams, this.charset, false);
+                formatQuery(sb, this.queryParams, this.charset,
+                        encodingPolicy == EncodingPolicy.ALL_RESERVED ? PercentCodec.UNRESERVED : PercentCodec.QUERY, false);
             } else if (this.query != null) {
                 sb.append("?");
                 PercentCodec.encode(sb, this.query, this.charset,
```

httpcore5/src/test/java/org/apache/hc/core5/net/TestURIBuilder.java
```diff
@@ -171,7 +171,7 @@ void testFormatQuery() {
     @Test
     void testHierarchicalUri() throws Exception {
         final URI uri = new URI("http", "stuff", "localhost", 80, "/some stuff", "param=stuff", "fragment");
-        final URIBuilder uribuilder = new URIBuilder(uri);
+        final URIBuilder uribuilder = new URIBuilder(uri).setEncodingPolicy(URIBuilder.EncodingPolicy.ALL_RESERVED);
```
ok2c, line 174: @arturobernalg I think this should no longer be necessary. `ALL_RESERVED` is default now.

httpcore5/src/test/java/org/apache/hc/core5/net/TestURIBuilder.java
```diff
@@ -472,6 +472,7 @@ void testFragmentEncoding() throws Exception {
                 .setScheme("https")
                 .setHost("somehost.com")
                 .setFragment("some fragment with all sorts of $tuff in it!!!")
+                .setEncodingPolicy(URIBuilder.EncodingPolicy.RFC_3986)
```
ok2c, line 475: @arturobernalg I think the test should also pass with `ALL_RESERVED`  default

httpcore5/src/test/java/org/apache/hc/core5/net/TestURIBuilder.java
```diff
@@ -617,7 +618,7 @@ void testHttpHost() throws Exception {
 
     @Test
     void testSetHostWithReservedChars() throws Exception {
-        final URIBuilder uribuilder = new URIBuilder();
+        final URIBuilder uribuilder = new URIBuilder().setEncodingPolicy(URIBuilder.EncodingPolicy.ALL_RESERVED);
```
ok2c, line 621: @arturobernalg I think this should no longer be necessary. `ALL_RESERVED` is default now.

ok2c (approved):

httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java
```diff
@@ -175,6 +203,22 @@ public URIBuilder setCharset(final Charset charset) {
...
     }
 
+    /**
+     * Sets the encoding policy for this {@link URIBuilder}.
+     * The encoding policy determines how URI components (e.g., query, fragment) are
+     * percent-encoded when building the URI string. If not set, the default policy
+     * is {@link EncodingPolicy#RFC_3986}.
+     *
+     * @param encodingPolicy the encoding policy to apply, or {@code null} to reset
+     *                       to the default ({@link EncodingPolicy#RFC_3986})
```
PascalSchumacher, line 213: I think this is no longer the case `ALL_RESERVED` seems to be the default now (see above).

httpcore5/src/test/java/org/apache/hc/core5/net/TestURIBuilder.java
arturobernalg: done

httpcore5/src/test/java/org/apache/hc/core5/net/TestURIBuilder.java
arturobernalg: done

httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java
arturobernalg: corrected

PR #518, opened 2025-03-14
HTTPCORE-778: URIBuilder to not encode sub-delims in URI fragment component for consistency with query component encoding

ok2c:


ok2c:
@arturobernalg Here's the fix for the stable 5.3.x branch. Please review.

PR #519, opened 2025-05-08
Httpclient 2368 - Support multiple TLS handshakes for HTTPS-proxy flows

arturobernalg:
HI @ok2c 
Not sure if this is a stupid idea or makes any sense,
but to solve HTTPCLIENT-2368 I’m proposing only a change in InternalDataChannel.startTls(...):

    Detect and clear any existing SSLIOSession, revert to raw I/O, then install a fresh SSLIOSession so startTls(...) can run twice (proxy → CONNECT → target) on the same channel without error.

Let me know if this narrow patch makes sense or if you’d prefer a different approach.


ok2c:
@arturobernalg The problem is not about executing TLS handshakes multiple times but about running multiple TLS sessions over the _same connection_ layered one over another.

PR #525, opened 2025-06-08
h2: add choosing framefactory impl

psmarko:
The FrameFactory interface is useful as, for example, it allows patching in PRIORITY support and changing frame values that are not generally configurable. It is set to the default impl in ClientH2StreamMultiplexerFactory making it hard to change. This PR adds the appropriate setter methods into ClientH2StreamMultiplexerFactory and H2RequesterBootstrap to choose a FrameFactory impl as so:
```java
H2RequesterBootstrap.bootstrap().setFrameFactory(impl).create()
```

### DIFF
--- a/httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/bootstrap/H2RequesterBootstrap.java
+++ b/httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/bootstrap/H2RequesterBootstrap.java
@@ -305,6 +305,11 @@ public class H2RequesterBootstrap {
         return this;
     }
 
+    /**
+     * Sets {@link FrameFactory} instance.
+     *
+     * @return this instance.
+     */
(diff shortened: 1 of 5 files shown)


httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/bootstrap/H2RequesterBootstrap.java
```diff
@@ -303,6 +305,16 @@ public final H2RequesterBootstrap setUriPatternType(final UriPatternType uriPatt
         return this;
     }
 
+    /**
```
ok2c, line 308: @psmarko This is a good start. Please add `@since 5.4`  tag to this method. There should also be a similar setter added to `H2MultiplexingRequesterBootstrap` and `H2ServerBootstrap`

ok2c (changes requested):

ok2c (approved):

PR #536, opened 2025-07-13
IOFunction functional interface

ok2c:
@arturobernalg I would like to move your `IOFunction` to core proper. Please double-check.


arturobernalg (approved):
@ok2c I'm OK with the change

PR #539, opened 2025-07-28
fix ContentType JavaDoc

winfriedgerlach:
The whole point of `ContentType.parseLenient()` is to **not** throw `UnsupportedCharsetException`. This should be reflected in the JavaDoc and the method signature.


ok2c:
@winfriedgerlach Thank you!

PR #540, opened 2025-07-28
Drop RxJava 2

rschmitt:
RxJava 2 is end-of-life as of February 28, 2021. `ReactiveTestUtils` can probably be dropped for 5.6 in favor of `Reactive3TestUtils`.


pom.xml
```diff
@@ -232,6 +231,7 @@
           <parameter>
             <breakBuildOnBinaryIncompatibleModifications>true</breakBuildOnBinaryIncompatibleModifications>
             <breakBuildOnSourceIncompatibleModifications>true</breakBuildOnSourceIncompatibleModifications>
+            <ignoreMissingClasses>true</ignoreMissingClasses>
```
ok2c, line 234: @rschmitt This is due to formatting, is it not?

rschmitt: No, I changed this by hand. This option appeared to be in the wrong section and wasn't taking effect. Without this option here, japicmp fails because the current version's classpath is missing the RxJava 2 classes (not to be confused with the missing `ReactiveTestUtils` class). Evidently, `ignoreMissingClasses` needs to be set in the build plugin config, not the reporting plugin config.

PR #541, opened 2025-07-31
HTTPCLIENT-2386: Fix TLS handshake timeout handling

arturobernalg:
Fix TLS handshake timeout handling ([HTTPCLIENT-2386](https://issues.apache.org/jira/browse/HTTPCLIENT-2386))

Use socketTimeout for TLS handshake and add dedicated timeout exception.

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/reactor/InternalDataChannel.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/reactor/InternalDataChannel.java
@@ -54,6 +54,7 @@ import org.apache.hc.core5.reactor.ssl.SSLMode;
 import org.apache.hc.core5.reactor.ssl.SSLSessionInitializer;
 import org.apache.hc.core5.reactor.ssl.SSLSessionVerifier;
 import org.apache.hc.core5.reactor.ssl.TlsDetails;
+import org.apache.hc.core5.reactor.ssl.TlsHandshakeTimeoutException;
 import org.apache.hc.core5.reactor.ssl.TransportSecurityLayer;
 import org.apache.hc.core5.util.Args;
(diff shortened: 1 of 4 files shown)


httpcore5/src/main/java/org/apache/hc/core5/reactor/InternalDataChannel.java
```diff
@@ -166,10 +167,17 @@ void onTimeout(final Timeout timeout) throws IOException {
         if (sessionListener != null) {
             sessionListener.timeout(currentSession);
         }
+
+        final SSLIOSession tlsSession = tlsSessionRef.get();
+        if (tlsSession != null && !tlsSession.isHandshakeComplete()) {
+            throw new TlsHandshakeTimeoutException("TLS handshake timed out after " + timeout);
```
garydgregory, line 173: @arturobernalg 
I think we should add a timeout only constructor and have the message in that constructor so it's normalized and not duplicated in each call site.

httpcore5/src/main/java/org/apache/hc/core5/reactor/InternalDataChannel.java
```diff
@@ -166,10 +167,17 @@ void onTimeout(final Timeout timeout) throws IOException {
...
+        }
+
         final IOEventHandler handler = ensureHandler(currentSession);
         handler.timeout(currentSession, timeout);
     }
 
+
```
garydgregory, line 180: Extra line?

arturobernalg:
> @arturobernalg I do not think we need this. The root cause of the problem is HTTPCLIENT-2386.

@ok2c This change is inspired by the discussion in  mailing list thread: https://lists.apache.org/thread/11bzsrztjm9y2862wo3x5ysoo9s1f5f7.

PR #542, opened 2025-08-06
Discontinue use of SocketSupport to set extended socket options

rschmitt:
The extended socket options related to TCP Keep-Alive were backported to Java 8 over five years ago. We can simply enable them directly without having to use reflection.

Related: https://github.com/apache/httpcomponents-client/pull/701


httpcore5/src/main/java/org/apache/hc/core5/io/SocketSupport.java
```diff
@@ -36,8 +36,11 @@
 
 /**
  * @since 5.3
+ *
+ * @deprecated No longer necessary, due to the backport of all supported ExtendedSocketOptions to Java 8
  */
 @Internal
+@Deprecated
```
ok2c: @rschmitt This class is internal and can just be dropped.

httpcore5/src/test/java/org/apache/hc/core5/io/TestSocketSupport.java
```diff
@@ -38,6 +38,7 @@
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 
+@SuppressWarnings("deprecation")
```
ok2c: @rschmitt Same. This test can be dropped now.

httpcore5/src/main/java/org/apache/hc/core5/io/SocketSupport.java
rschmitt: What's more, I don't think we've ever released a version of httpclient that depended on this class, so this is the perfect time to delete this class!

PR #544, opened 2025-08-08
Check for keep-alive option support

rschmitt:
It turns out that keep-alive options are supported on all modern combinations of Java runtimes and operating systems, with the sole excpetion of Java 8 on Windows. The keep-alive options are specifically unavailable on this combination: you can refer to them in code, but attempting to set them will result in an UnsupportedOperationException.

The good news is that there's a method we can call, `Sockets.supportedOptions()`, that returns a complete list of socket options supported by the current runtime. This change simply uses that functionality to check for support before setting keep-alive options on any socket.


ok2c (approved):

httpcore5/src/main/java/org/apache/hc/core5/util/ReflectionUtils.java
```diff
@@ -89,4 +97,7 @@ public static int determineJRELevel() {
         return 7;
     }
 
+    public static boolean supportsKeepAliveOptions() {
```
garydgregory, line 100: Missing Javadoc. Adding this here seems misleading since it's not related to reflection. Isn't there a better place?

rschmitt: It's not _literally_ reflection, but it is introspection of the current runtime capabilities, so I figured this was a close enough match and would save me the trouble of creating a new `Utils` class.

PR #547, opened 2025-08-10
Stale check command for async connections

ok2c:
@rschmitt This change-set introduces a relatively cheap 'stale' connection check command that works with both HTTP/1.1 and H2 protocols and can be used instead of a more expensive Ping command.

Please take a look.


httpcore5/src/main/java/org/apache/hc/core5/http/nio/command/StaleCheckCommand.java
```diff
@@ -0,0 +1,58 @@
...
+        this.callback = Args.notNull(callback, "Callback");
+    }
+
+    public FutureCallback<Boolean> getCallback() {
+        return callback;
+    }
+
+    @Override
+    public boolean cancel() {
+        return true;
```
rschmitt, line 59: Cancel the callback

httpcore5/src/main/java/org/apache/hc/core5/http/impl/nio/AbstractHttp1StreamDuplexer.java
```diff
@@ -429,6 +433,20 @@ void requestShutdown(final CloseMode closeMode) {
         ioSession.setEvent(SelectionKey.OP_WRITE);
     }
 
+    void doStalecheck(final FutureCallback<Boolean> callback) throws IOException {
+        if (!ioSession.isOpen() || connState.compareTo(ConnectionState.ACTIVE) > 0) {
+            callback.completed(false);
+        }
+        final ByteBuffer buffer = ByteBuffer.allocate(0);
+        final int bytesRead = ioSession.channel().read(buffer);
```
rschmitt: I think this read is actually redundant. `Command` is modeled as a write event, and reads are processed before writes. So if the connection is closed, we'll already know by the time we get here, _provided_ that we include the changes from https://github.com/apache/httpcomponents-core/pull/543.

The end result ends up being very similar to the "synchronization barrier" between the event loop and the connection pool that I was talking about in that PR. The internal race condition basically goes away as long as connection reuse is completed through the event loop, which then has a chance to update all the relevant bookkeeping with respect to whatever IO events are pending.

rschmitt (changes requested):

rschmitt:
I made the following changes locally:
1. I pulled in this change
2. I added https://github.com/apache/httpcomponents-core/pull/543 (specifically dfa2cd51c038f5f3259d10a5cfba73953245f188)
3. I taught `PoolingAsyncClientConnectionManager` to submit a `StaleCheckCommand` as the implementation of validateAfterInactivity on HTTP/1.1 connections
4. I set `setValidateAfterInactivity(ZERO_MILLISECONDS)` in `TestConnectionClosureRace`
When I do all of this, the results are dramatic:

```
Http: Sequential requests (rapid): 2,500 succeeded; 0 failed (100.00% success rate)
Http: Sequential requests (slow): 10 succeeded; 0 failed (100.00% success rate)
Http: Single large batch: 30 succeeded; 0 failed (100.00% success rate)
Http: Multiple small batches: 15 succeeded; 0 failed (100.00% success rate)

Https: Sequential requests (rapid): 2,499 succeeded; 1 failed (99.96% success rate, 0.00% retriable)
Https: Sequential requests (slow): 10 succeeded; 0 failed (100.00% success rate)
Https: Single large batch: 29 succeeded; 1 failed (96.67% success rate, 0.00% retriable)
Https: Multiple small batches: 13 succeeded; 2 failed (86.67% success rate, 0.00% retriable)
```

When I disable inactive connection validation, I get the same results I've been getting:

```
Http: Sequential requests (rapid): 2,494 succeeded; 6 failed (99.76% success rate, 0.24% retriable)
Http: Sequential requests (slow): 10 succeeded; 0 failed (100.00% success rate)
Http: Single large batch: 15 succeeded; 15 failed (50.00% success rate, 50.00% retriable)
Http: Multiple small batches: 10 succeeded; 5 failed (66.67% success rate, 33.33% retriable)

Https: Sequential requests (rapid): 2,476 succeeded; 24 failed (99.04% success rate, 0.96% retriable)
Https: Sequential requests (slow): 10 succeeded; 0 failed (100.00% success rate)
Https: Single large batch: 15 succeeded; 15 failed (50.00% success rate, 50.00% retriable)
Https: Multiple small batches: 10 succeeded; 5 failed (66.67% success rate, 26.67% retriable)
```

We're definitely on the right track. If I'm right about the IO in `doStalecheck`, that code can probably all be deleted, and `StaleCheckCommand` can be renamed `ConnectionLeaseCommand` or something.

rschmitt:
> I think there's might be a race condition here involving command execution and connection closure.

I think I was mistaken about this. The actual issue might have been the no-op implementation of `StaleCheckCommand::cancel`. `IOSessionImpl::enqueue` enqueues the command, but then cancels it if `isStatusClosed()`, which seems like it should be okay.

ok2c:
> We're definitely on the right track. If I'm right about the IO in `doStalecheck`, that code can probably all be deleted, and `StaleCheckCommand` can be renamed `ConnectionLeaseCommand` or something.

@rschmitt This looks encouraging.

rschmitt:
https://github.com/rschmitt/httpcomponents-core/tree/stale_conn_check_async

arturobernalg (approved):
@ok2c  LGTM

rschmitt:
> This should be the case already. Connections should immediately go into graceful shutdown state immediately upon the receipt of GOAWAY. I will double-check. though

It'll take me a few hours to add HTTP/2 support to my tester. I'll follow up with that later, as well as with the conversion of `TestConnectionClosureRace` into an example class.

ok2c:
> Please incorporate the changes from [dfa2cd5](https://github.com/apache/httpcomponents-core/commit/dfa2cd51c038f5f3259d10a5cfba73953245f188) (or reopen #543 and ship them separately)

@rschmitt I am not in favor of re-opening and committing #543 for the reasons stated already. I think the removal of the read operation from the stale connection check was a mistake.

rschmitt:
@ok2c I tested with 1bf385947bb8a2b5752a2403040de73b74365e07 and the race condition still exists on HTTP (but is fixed on HTTPS). If you don't want to ship #543 then I think reintroducing the read operation will work.

ok2c:
>  I think reintroducing the read operation will work

@rschmitt That is unfortunate because https://github.com/apache/httpcomponents-core/commit/1bf385947bb8a2b5752a2403040de73b74365e07 already does an extra read. It can be though the call is ignored due to the read buffer capacity being zero.

ok2c:
@rschmitt What do you want me to do about this one in the meantime? I am not sure there is anything else I could do to improve the check. Do you want me to commit the change-set as is or do you want me to drop it altogether?

PR #552, opened 2025-08-22
HTTP/2: implement RFC 9218 Extensible Prioritization Scheme for HTTP.

arturobernalg:
This PR adds core support for RFC 9218: we advertise SETTINGS_NO_RFC7540_PRIORITIES=1, gate PRIORITY_UPDATE emission on the peer’s setting (or pre-SETTINGS), parse/accept PRIORITY_UPDATE frames in the base muxer, and introduce  PriorityParamsParser for the Priority header (u/i members, tolerant parsing).

### DIFF
--- a/httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/AbstractH2StreamMultiplexer.java
+++ b/httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/AbstractH2StreamMultiplexer.java
@@ -31,10 +31,13 @@ import java.net.SocketAddress;
 import java.nio.BufferOverflowException;
 import java.nio.ByteBuffer;
 import java.nio.channels.SelectionKey;
+import java.nio.charset.StandardCharsets;
 import java.util.Deque;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
 import java.util.Queue;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedDeque;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -66,6 +69,7 @@ import org.apache.hc.core5.http.nio.command.RequestExecutionCommand;
 import org.apache.hc.core5.http.nio.command.ShutdownCommand;
 import org.apache.hc.core5.http.protocol.HttpContext;
 import org.apache.hc.core5.http.protocol.HttpProcessor;
+import org.apache.hc.core5.http.HttpHeaders;
 import org.apache.hc.core5.http2.H2ConnectionException;
 import org.apache.hc.core5.http2.H2Error;
 import org.apache.hc.core5.http2.H2StreamResetException;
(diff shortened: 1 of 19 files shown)


ok2c:
@arturobernalg Before we can do anything of this kind we must ensure conformance with [RFC 9113](https://datatracker.ietf.org/doc/html/rfc9113) first.

httpcore5-h2/src/main/java/org/apache/hc/core5/http2/priority/PriorityParamsParser.java
```diff
@@ -0,0 +1,136 @@
...
+
+    public static PriorityParams parse(final String headerValue) {
+        Integer urgency = null;     // null means “u absent”
+        Boolean incremental = null; // null means “i absent”
+
+        if (headerValue == null || headerValue.isEmpty()) {
+            return new PriorityParams(null, null);
+        }
+
+        final CharArrayBuffer buf = new CharArrayBuffer(headerValue.length());
```
ok2c: @arturobernalg Please do not do that. There is no need to create a copy of the header value. The parser can work with any `CharSequence`.  Can't you re-use `MessageSupport` parsing routines here?

Otherwise, looks good.

ok2c (changes requested):

httpcore5-h2/src/main/java/org/apache/hc/core5/http2/priority/PriorityParamsParser.java
arturobernalg: @ok2c  please do another pass

ok2c:
@arturobernalg Please also add RFC 9218 to the protocol support section in `README.md`

arturobernalg:
> @arturobernalg Please also add RFC 9218 to the protocol support section in `README.md`

@ok2c done

httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/AbstractH2StreamMultiplexer.java
```diff
@@ -1334,6 +1359,38 @@ H2Stream createStream(final H2StreamChannel channel, final H2StreamHandler strea
         return streams.createActive(channel, streamHandler);
     }
 
+    public final void sendPriorityUpdate(final int prioritizedStreamId, final PriorityValue value) throws IOException {
```
ok2c, line 1362: @arturobernalg By the way this method does not seem to be used anywhere. Is this intended?

ok2c:
@arturobernalg I did a sloppy job reviewing your code. It needs more work.

httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/AbstractH2StreamMultiplexer.java
arturobernalg: @ok2c Yes, that was intentional. The logic was inlined into the stream submit path, so the standalone helper became redundant. I'm going to removed it.

ok2c: @arturobernalg Please also remove `H2RequestPriority` interceptor for now. It is not being used anywhere in core and I presume your intention was to use it in client. In this case the interceptor should be in client.

ok2c: @arturobernalg Also `PriorityFormatter` is not being used anywhere in core. If they are not used in core they should not be in core. Re-submit them in client. Believe me you do not want to depend on a new release of core to be able to fix a bug in _client_.

ok2c: @arturobernalg Please let me know if you can live with #567. The functionality should be moved to client and we should put more work into the API design

PR #553, opened 2025-09-08
H2 transport conformant with RFC 9113 regarding

arturobernalg:
It introduces directional enforcement inside AbstractH2StreamMultiplexer:

Locally-initiated streams are gated by the peer’s advertised MAX_CONCURRENT_STREAMS.
Remotely-initiated streams exceeding our advertised limit are immediately refused with RST_STREAM(REFUSED_STREAM).
The same inbound check applies to PUSH_PROMISE promised streams.
After applying remote SETTINGS, we request session output so queued commands progress when the peer raises the limit.

Implementation is contained to the multiplexer (no new controller class). Active stream counting ignores terminated streams. Flow control logic remains untouched.TREAM) Add tests for inbound headers, outbound gating, and push-promise refusal; nudge queued opens on SETTINGS ACK

### DIFF
--- a/httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/AbstractH2StreamMultiplexer.java
+++ b/httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/AbstractH2StreamMultiplexer.java
@@ -528,7 +528,8 @@ abstract class AbstractH2StreamMultiplexer implements Identifiable, HttpConnecti
         }
 
         if (connState.compareTo(ConnectionHandshake.ACTIVE) <= 0 && remoteSettingState == SettingsHandshake.ACKED) {
-            while (streamMap.size() < remoteConfig.getMaxConcurrentStreams()) {
+            final int outboundLimit = remoteConfig.getMaxConcurrentStreams();
+            while (countActiveLocalInitiated() < outboundLimit) {
                 final Command command = ioSession.poll();
                 if (command == null) {
                     break;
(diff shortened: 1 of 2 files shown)


httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/AbstractH2StreamMultiplexer.java
```diff
@@ -1456,6 +1497,9 @@ public void push(final List<Header> headers, final AsyncPushProducer pushProduce
 
             ioSession.getLock().lock();
             try {
+                if (countActiveLocalInitiated() >= remoteConfig.getMaxConcurrentStreams()) {
```
ok2c: @arturobernalg Imagine being a user who wants to push 10 messages to the client being completely unaware of the MAX_CONCURRENT_STREAMS limit imposed by the client and getting some of those messages dropped with some obscure  "Outbound stream limit reached" message. What would be your reaction?

arturobernalg: @ok2c  I reworked the patch so we never drop user traffic because of the peer’s MAX_CONCURRENT_STREAMS.

ok2c: @arturobernalg so, the `MAX_CONCURRENT_STREAMS` does not apply to the server pushed streams, as before?

arturobernalg: > @arturobernalg so, the `MAX_CONCURRENT_STREAMS` does not apply to the server pushed streams, as before?

@ok2c You’re right to call this out. MAX_CONCURRENT_STREAMS does apply to server-initiated streams.  my bad

ok2c: > > @arturobernalg so, the `MAX_CONCURRENT_STREAMS` does not apply to the server pushed streams, as before?
> 
> @ok2c You’re right to call this out. MAX_CONCURRENT_STREAMS does apply to server-initiated streams. my bad

@arturobernalg You are trying to brute force a fix on an API that simply does not have enough flexibility to support the protocol requirements. As I said before a proper solution to this issue is not possible without an extensive refactoring of the H2 protocol handling code.

PR #554, opened 2025-09-10
H2: enforce inbound concurrency at activation

arturobernalg:
Refactor inbound stream lifecycle to be RFC-conformant: reservation (PUSH_PROMISE) vs activation (first HEADERS). Enforce the inbound concurrency limit atomically at activation; refuse overflow with RST_STREAM(REFUSED_STREAM) while draining any outstanding header block.

@ok2c maybe this can work.


ok2c:
@arturobernalg I am sorry this change-set came while I was working on the same problem. I ended up with a different solution  that takes  two atomic integers only instead of two concurrent sets to keep track of active H2 streams. It should be somewhat more efficient but admittedly took a lot of effort on my part to get right. And it also works with both inbound and outbound streams.

Please review my code in master and submit any RFC conformance improvements I may have missed.

PR #555, opened 2025-09-11
Feature/hc5 serverbootstrap convenience methods

cdw8j:
I have added two convenience methods to the ServerBootstrap that allow simple customization of the RequestRouter in the HttpServer (custom server info string and switching LOCAL_AUTHORITY_RESOLVER scheme). While it is possible to create an HttpServer with such a custom RequestRouter within the application, this results in really messy and complicated code, which can easily be spared with these very simple convenience methods.

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/http/impl/bootstrap/ServerBootstrap.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/http/impl/bootstrap/ServerBootstrap.java
@@ -98,10 +98,12 @@ public class ServerBootstrap {
     private HttpConnectionFactory<? extends DefaultBHttpServerConnection> connectionFactory;
     private ExceptionListener exceptionListener;
     private Http1StreamListener streamListener;
+    private boolean localAuthorityResolver;
 
     private ServerBootstrap() {
         this.routeEntries = new ArrayList<>();
         this.filters = new ArrayList<>();
+        this.localAuthorityResolver = false;
     }
 
     public static ServerBootstrap bootstrap() {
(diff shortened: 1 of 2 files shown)


httpcore5/src/main/java/org/apache/hc/core5/http/impl/bootstrap/ServerBootstrap.java
```diff
@@ -182,6 +185,17 @@ public final ServerBootstrap setResponseFactory(final HttpResponseFactory<Classi
         return this;
     }
 
+
+    public final ServerBootstrap setServerInfo(final String serverInfo) {
+        this.serverInfo = serverInfo;
+        return this;
+    }
+
+    public final ServerBootstrap setLocalAuthResolver() {
```
garydgregory: This one is really not worth the API clutter IMO.

cdw8j: I agree the setServerInfo() is arguable, but the setLocalAuthResolver() really makes life easier on application side, because it avoids the clutter related to creating a custom RequestRouter. Should I create a pull request only for the setLocalAuthResolver()?

cdw8j: Removed the `setServerInfo()` from pull request

httpcore5/src/main/java/org/apache/hc/core5/http/impl/bootstrap/ServerBootstrap.java
```diff
@@ -346,6 +348,16 @@ public final ServerBootstrap addFilterLast(final String name, final HttpFilterHa
         return this;
     }
 
+    /**
+     * Create {@link RequestRouter} with LOCAL_AUTHORITY_RESOLVER (default: IGNORE_PORT_AUTHORITY_RESOLVER).
+     *
+     * @since 5.3.6
+     */
+    public final ServerBootstrap enableLocalAuthorityResolver() {
```
ok2c, line 356: @cdw8j `#useLocalAuthorityResolver` maybe?

PR #556, opened 2025-09-12
added useLocalAuthorityResolver() to ServerBootstrap and AsyncServerBootStrap classes

cdw8j:
(Follow-up to pull request #555 which I messed up while trying to rebase it from 5.3.x to master.)

I have added a convenience method to the ServerBootstrap and AsyncServerBootstrap classes that configures the RequestRouter to use the LOCAL_AUTHORITY_RESOLVER scheme. While it is possible to create an HttpServer with such a custom RequestRouter within the application, this results in really messy and complicated code, which can easily be spared with this very simple convenience method.

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/http/impl/bootstrap/AsyncServerBootstrap.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/http/impl/bootstrap/AsyncServerBootstrap.java
@@ -95,6 +95,7 @@ public class AsyncServerBootstrap {
     private IOSessionListener sessionListener;
     private Http1StreamListener streamListener;
     private IOReactorMetricsListener threadPoolListener;
+    private boolean localAuthorityResolver;
 
     private AsyncServerBootstrap() {
         this.routeEntries = new ArrayList<>();
@@ -435,6 +436,16 @@ public class AsyncServerBootstrap {
         return this;
     }
 
+    /**
+     * Create {@link RequestRouter} with LOCAL_AUTHORITY_RESOLVER (default: IGNORE_PORT_AUTHORITY_RESOLVER).
+     *
+     * @since 5.4
+     */
+    public final AsyncServerBootstrap useLocalAuthorityResolver(final boolean localAuthorityResolver) {
+        this.localAuthorityResolver = localAuthorityResolver;
+        return this;
+    }
+
     public HttpAsyncServer create() {
         final String actualCanonicalHostName = canonicalHostName != null ? canonicalHostName : InetAddressUtils.getCanonicalLocalHostName();
         final HttpRequestMapper<Supplier<AsyncServerExchangeHandler>> requestRouterCopy;
@@ -451,9 +462,9 @@ public class AsyncServerBootstrap {
                 requestRouterCopy = requestRouter;
             } else {
                 requestRouterCopy = RequestRouter.create(
-                        new URIAuthority(actualCanonicalHostName),
+                        this.localAuthorityResolver ? RequestRouter.LOCAL_AUTHORITY : new URIAuthority(actualCanonicalHostName),
                         UriPatternType.URI_PATTERN, routeEntries,
-                        RequestRouter.IGNORE_PORT_AUTHORITY_RESOLVER,
+                        this.localAuthorityResolver ? RequestRouter.LOCAL_AUTHORITY_RESOLVER : RequestRouter.IGNORE_PORT_AUTHORITY_RESOLVER,
                         requestRouter);
             }
         }
--- a/httpcore5/src/main/java/org/apache/hc/core5/http/impl/bootstrap/ServerBootstrap.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/http/impl/bootstrap/ServerBootstrap.java
@@ -98,6 +98,7 @@ public class ServerBootstrap {
     private HttpConnectionFactory<? extends DefaultBHttpServerConnection> connectionFactory;
     private ExceptionListener exceptionListener;
     private Http1StreamListener streamListener;
+    private boolean localAuthorityResolver;
 
     private ServerBootstrap() {
         this.routeEntries = new ArrayList<>();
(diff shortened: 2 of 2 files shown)


httpcore5/src/main/java/org/apache/hc/core5/http/impl/bootstrap/AsyncServerBootstrap.java
```diff
@@ -95,10 +95,12 @@ public class AsyncServerBootstrap {
     private IOSessionListener sessionListener;
     private Http1StreamListener streamListener;
     private IOReactorMetricsListener threadPoolListener;
+    private boolean localAuthorityResolver;
 
     private AsyncServerBootstrap() {
         this.routeEntries = new ArrayList<>();
         this.filters = new ArrayList<>();
+        this.localAuthorityResolver = false;
```
garydgregory: Don't initialize a primitive to its default value.

httpcore5/src/main/java/org/apache/hc/core5/http/impl/bootstrap/AsyncServerBootstrap.java
```diff
@@ -435,6 +437,16 @@ public final AsyncServerBootstrap addFilterLast(final String name, final AsyncFi
         return this;
     }
 
+    /**
+     * Create {@link RequestRouter} with LOCAL_AUTHORITY_RESOLVER (default: IGNORE_PORT_AUTHORITY_RESOLVER).
+     *
+     * @since 5.4
+     */
+    public final AsyncServerBootstrap useLocalAuthorityResolver() {
```
garydgregory: Bad API design IMO, pass in the the boolean value, such that, for example, an event like a UI checkbox (or anything really) can toggle the value.

httpcore5/src/main/java/org/apache/hc/core5/http/impl/bootstrap/ServerBootstrap.java
```diff
@@ -346,6 +348,16 @@ public final ServerBootstrap addFilterLast(final String name, final HttpFilterHa
         return this;
     }
 
+    /**
+     * Create {@link RequestRouter} with LOCAL_AUTHORITY_RESOLVER (default: IGNORE_PORT_AUTHORITY_RESOLVER).
+     *
+     * @since 5.4
+     */
+    public final ServerBootstrap useLocalAuthorityResolver() {
+        this.localAuthorityResolver = true;
```
garydgregory: Same comment as above.

httpcore5/src/main/java/org/apache/hc/core5/http/impl/bootstrap/ServerBootstrap.java
```diff
@@ -98,10 +98,12 @@ public class ServerBootstrap {
     private HttpConnectionFactory<? extends DefaultBHttpServerConnection> connectionFactory;
     private ExceptionListener exceptionListener;
     private Http1StreamListener streamListener;
+    private boolean localAuthorityResolver;
 
     private ServerBootstrap() {
         this.routeEntries = new ArrayList<>();
         this.filters = new ArrayList<>();
+        this.localAuthorityResolver = false;
```
garydgregory: Same comment as above.

cdw8j:
Hello @garydgregory,

seems this request is in deadlock, so let me explain the background of my request.

I migrated a project from httpcore4 to httpcore5 just recently and when it didn't work, I had to dive into the httpcore5 sources to understand what was going wrong and that I need to create a custom RequestRouter with LOCAL_AUTHORITY_RESOLVER scheme to get the same behavior as before with httpcore4 just using simple ServerBootstrap methods. I might be the last on earth migrating an ancient httpcore4 application, but in this scenario and I guess also in others where the host name shall be irrelevant (and which is the reason why the LOCAL_AUTHORITY_RESOLVER scheme exists in httpcore5), it is quite a pain IMO to build the custom RequestRouter in the application itself, especially because this isn't really very well documented.

But in the end, it's not my decision, so if you decide it's not worth the API extension, then just reject the pull request and nobody will get hurt either.

httpcore5/src/main/java/org/apache/hc/core5/http/impl/bootstrap/AsyncServerBootstrap.java
cdw8j: Changed this in my revised commit

httpcore5/src/main/java/org/apache/hc/core5/http/impl/bootstrap/AsyncServerBootstrap.java
cdw8j: Changed this in my revised commit

httpcore5/src/main/java/org/apache/hc/core5/http/impl/bootstrap/ServerBootstrap.java
cdw8j: Changed this in my revised commit

httpcore5/src/main/java/org/apache/hc/core5/http/impl/bootstrap/ServerBootstrap.java
cdw8j: Changed this in my revised commit

httpcore5/src/main/java/org/apache/hc/core5/http/impl/bootstrap/ServerBootstrap.java
```diff
@@ -346,6 +347,16 @@ public final ServerBootstrap addFilterLast(final String name, final HttpFilterHa
         return this;
     }
 
+    /**
+     * Create {@link RequestRouter} with LOCAL_AUTHORITY_RESOLVER (default: IGNORE_PORT_AUTHORITY_RESOLVER).
```
ok2c, line 351: @cdw8j  Please do not mention `LOCAL_AUTHORITY_RESOLVER` and  `IGNORE_PORT_AUTHORITY_RESOLVER`. This is an implementation details and it should be mentioned in javadocs

httpcore5/src/main/java/org/apache/hc/core5/http/impl/bootstrap/ServerBootstrap.java
```diff
@@ -346,6 +347,16 @@ public final ServerBootstrap addFilterLast(final String name, final HttpFilterHa
...
+     *
+     * @since 5.4
+     */
+    public final ServerBootstrap useLocalAuthorityResolver(final boolean localAuthorityResolver) {
```
ok2c, line 355: @cdw8j Please use a better name for the method. `#routeToLocalAuthority` for example.

ok2c (changes requested):
@cdw8j My apologies. I actually did not submit my review last time.

Please see my comments,

ok2c:
> What do you think? Should I create a separate pull request for this option and you decide which is preferrable?

@cdw8j I personally like this idea much better. If you do it fast enough there is a a good chance it will make it into the coming 5.4-alpha1 release

cdw8j:
@ok2c not too much of a challenge :) see [pull request #565 ](https://github.com/apache/httpcomponents-core/pull/565)

PR #558, opened 2025-09-12
Bug fix: Classic RequestListener to continue accepting incoming connections in case of an unexpected exception caused by a single connection

cdw8j:
The RequestListener thread terminates when an exception occurs during connection creation or in the worker thread. Specifically, I have noticed this issue simply when an http connection is attempted to an HttpServer with SSLContext - the server essentially terminates.

So instead of throwing the corresponding exception, it should be handled by the application's exceptionListener.

(This is also how it was implemented in httpcore4.)


ok2c:
@cdw8j Cherry-picked to master

PR #564, opened 2025-10-02
HTTPCORE: Add RFC 7639 ALPN header codec and constant

arturobernalg:
New public `AlpnHeader` implementing RFC7639.

Add HttpHeaders.ALPN. Canonical encoding: keep tchar except %; uppercase hex; UTF-8; liberal parse (OWS, lower-case %hh).

Note: A follow-up PR for httpclient5 will inject ALPN on CONNECT (classic & async) via an optional provider


httpcore5/src/main/java/org/apache/hc/core5/http/message/AlpnHeader.java
```diff
@@ -0,0 +1,230 @@
...
+ * Liberal parsing: accepts lowercase hex in {@code %hh}; leaves malformed or incomplete
+ * percent sequences literal. Quoted values are not produced by this codec and are not required
+ * by the spec (ALPN IDs are tokens).
+ *
+ * <p><strong>Thread-safety:</strong> This class is immutable and thread-safe.</p>
+ *
+ * @since 5.4
+ */
+@Contract(threading = ThreadingBehavior.IMMUTABLE)
+public final class AlpnHeader {
```
ok2c: @arturobernalg It is going to be a support class used in client, right? Maybe it should be called `AlpnHeaderSupport` and be ` @Internal`?

httpcore5/src/main/java/org/apache/hc/core5/http/message/AlpnHeaderSupport.java
```diff
@@ -0,0 +1,230 @@
...
+
+    /**
+     * Parses an {@code ALPN} header value into decoded protocol IDs.
+     * Uses {@link Tokenizer} to handle OWS and commas without intermediate copies.
+     *
+     * @param value header value (may be {@code null} or empty)
+     * @return decoded IDs; never {@code null}
+     * @since 5.4
+     */
+    public static List<String> parseValue(final String value) {
```
ok2c, line 74: @arturobernalg Use `MessageSupport#parseElementList` or `MessageSupport#parseTokens` here.

httpcore5/src/main/java/org/apache/hc/core5/http/message/AlpnHeader.java
arturobernalg: @ok2c I’m working on the client-side part now.

arturobernalg:
> @arturobernalg I will commit the header constant but codecs should go into client. This feature belongs there.

@ok2c I'll the proper changes and I will moved to the client

PR #567, opened 2025-10-04
Removed unused H2 priority related code. To be moved to client

ok2c:


ok2c:
@arturobernalg Okay. here's the latest

1. I removed unused protocol code
2. I completely re-wrote `PriorityFormatter` for memory efficiency. Please take a look at my implementation and yours and compare them in terms of intermediate garbage generation. Message parsing / formatting usually used are in the hot execution path and _MUST_ efficient. That also implies memory efficiency. 
3. Changes the sample application. Classic over Async is still experimental and will likely be experimental for a long time. we should not promote it unless `InputStream` / `OutputStream` compatibility is really a necessity.

PR #578, opened 2025-11-14
Add pooled ByteBuffer allocator with size classes

arturobernalg:
This is the first step towards a pluggable pooled ByteBuffer allocator.
The patch adds PooledByteBufferAllocator (power-of-two size buckets, global pool + per-thread caches) and switches the HTTP/2 FrameFactory and the benchmark to use ByteBufferAllocator. Behaviour is unchanged except for using pooled buffers for small control frames.
If this direction looks reasonable, I’ll follow up by threading the allocator through IOSession/SSLIOSession and the async codecs and add minimal metrics; otherwise I’ll keep it local to HTTP/2.

### ByteBuffer allocator throughput (JMH)

| Benchmark                                      | bufferSize | iterations | Mode  | Cnt | Score (ops/ms) | Error (ops/ms) |
|-----------------------------------------------|-----------:|-----------:|:------|----:|---------------:|---------------:|
| pooled_allocator_shared                       |       1024 |        100 | thrpt |  10 |       1644.982 |         14.006 |
| pooled_allocator_shared                       |       8192 |        100 | thrpt |  10 |        533.638 |         34.307 |
| pooled_allocator_shared                       |      65536 |        100 | thrpt |  10 |         59.422 |          0.937 |
| pooled_allocator_thread_local                 |       1024 |        100 | thrpt |  10 |        539.612 |          5.518 |
| pooled_allocator_thread_local                 |       8192 |        100 | thrpt |  10 |        201.345 |          4.451 |
| pooled_allocator_thread_local                 |      65536 |        100 | thrpt |  10 |         19.603 |          0.501 |
| simple_allocator_shared                       |       1024 |        100 | thrpt |  10 |        172.750 |          4.893 |
| simple_allocator_shared                       |       8192 |        100 | thrpt |  10 |         23.083 |          0.199 |
| simple_allocator_shared                       |      65536 |        100 | thrpt |  10 |          2.883 |          0.037 |
| simple_allocator_thread_local                 |       1024 |        100 | thrpt |  10 |        129.873 |          1.075 |
| simple_allocator_thread_local                 |       8192 |        100 | thrpt |  10 |         21.075 |          0.088 |
| simple_allocator_thread_local                 |      65536 |        100 | thrpt |  10 |          2.401 |          0.062 |

@ok2c WDYT?

### DIFF
--- a/httpcore5-testing/pom.xml
+++ b/httpcore5-testing/pom.xml
@@ -171,6 +171,9 @@
                     <argument>-classpath</argument>
                     <classpath/>
                     <argument>org.openjdk.jmh.Main</argument>
+                    <!-- GC profiler -->
+                    <argument>-prof</argument>
+                    <argument>gc</argument>
                     <argument>-rf</argument>
                     <argument>json</argument>
                     <argument>-rff</argument>
--- a/httpcore5-testing/src/test/java/org/apache/hc/core5/benchmark/ByteBufferAllocatorBenchmark.java
+++ b/httpcore5-testing/src/test/java/org/apache/hc/core5/benchmark/ByteBufferAllocatorBenchmark.java
@@ -1,0 +1,202 @@
+/*
+ * ====================================================================
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ * ====================================================================
+ *
+ * This software consists of voluntary contributions made by many
+ * individuals on behalf of the Apache Software Foundation.  For more
+ * information on the Apache Software Foundation, please see
+ * <http://www.apache.org/>.
+ *
+ */
+package org.apache.hc.core5.benchmark;
+
+import java.nio.ByteBuffer;
+import java.util.concurrent.TimeUnit;
+
+import org.apache.hc.core5.io.ByteBufferAllocator;
+import org.apache.hc.core5.io.PooledByteBufferAllocator;
+import org.apache.hc.core5.io.SimpleByteBufferAllocator;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Fork;
+import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Param;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.Threads;
+import org.openjdk.jmh.annotations.Warmup;
+import org.openjdk.jmh.infra.Blackhole;
+
+@BenchmarkMode(Mode.Throughput)
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+@Warmup(iterations = 5, time = 1, timeUnit = TimeUnit.SECONDS)
+@Measurement(iterations = 10, time = 1, timeUnit = TimeUnit.SECONDS)
+@Fork(1)
+public class ByteBufferAllocatorBenchmark {
+
+    /**
+     * Per-thread state: each thread gets its own allocators.
+     * This measures the best-case hot-path behaviour with no contention
+     * on the pooled allocator.
+     */
+    @State(Scope.Thread)
+    public static class ThreadLocalState {
+
+        @Param({"1024", "8192", "65536"})
+        public int bufferSize;
+
+        @Param({"100"})
+        public int iterations;
+
+        public ByteBufferAllocator simpleAllocator;
+        public ByteBufferAllocator pooledAllocator;
+
+        public byte[] payload;
+
+        @Setup
+        public void setUp() {
+            this.simpleAllocator = SimpleByteBufferAllocator.INSTANCE;
+            this.pooledAllocator = new PooledByteBufferAllocator(
+                    1024,        // min pooled size
+                    256 * 1024,  // max pooled size
+                    1024,        // maxGlobalPerBucket
+                    64           // maxLocalPerBucket
+            );
+            this.payload = new byte[bufferSize / 2];
+            for (int i = 0; i < this.payload.length; i++) {
+                this.payload[i] = (byte) (i & 0xFF);
+            }
+        }
+
+    }
+
+    /**
+     * Shared state: all threads share the same allocators.
+     * This measures contention on the global buckets and any shared
+     * structures inside the pooled allocator.
+     */
+    @State(Scope.Benchmark)
+    public static class SharedState {
+
+        @Param({"1024", "8192", "65536"})
+        public int bufferSize;
+
+        @Param({"100"})
+        public int iterations;
+
+        public ByteBufferAllocator simpleAllocator;
+        public ByteBufferAllocator pooledAllocator;
+
+        public byte[] payload;
+
+        @Setup
+        public void setUp() {
+            this.simpleAllocator = SimpleByteBufferAllocator.INSTANCE;
+            this.pooledAllocator = new PooledByteBufferAllocator(
+                    1024,
+                    256 * 1024,
+                    1024,
+                    64
+            );
+            this.payload = new byte[bufferSize / 2];
+            for (int i = 0; i < this.payload.length; i++) {
+                this.payload[i] = (byte) (i & 0xFF);
+            }
+        }
+
+    }
+
+    // --------- Per-thread allocators (your original semantics) ---------
+
+    @Benchmark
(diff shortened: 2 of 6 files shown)


rschmitt:
I asked [Aleksey Shipilëv](https://shipilev.net/) for his thoughts:
> Depends. In a pure allocation benchmark, allocation would likely be on par with reuse. But once you get far from that ideal, awkward things start to happen.
> 1. When there is _any_ non-trivial live set in the heap, GC would have to at least visit it every so often; that "so often" is driven by GC frequency, which is driven by allocation rate. Pure allocation speed and pure reclamation cost becomes much less relevant in this scenario -- what _else_ is happenning dominates hard. Generational GCs win you some, but they really only prolong the inevitable.
> 2. When objects are allocated, they are nominally zeroed. Under high allocation rate, that is easily the slowest part, think ~10 GB/sec per thread. Re-use often comes with avoiding these cleanups, often at the cost of weaker security posture (leaking data between reused buffers).
> 3. For smaller objects, the metadata management (headers, all that fluff) dominates the allocation path performance, _and_ is often logically intermixed with the _real_ work. E.g. you rarely allocate 10M objects just because, there is likely some compute in between. But allocating `new byte[BUF_SIZE]` (`BUF_SIZE=1M` defined in another file) is very easy. So hitting (1) and (2) is much easier the larger the object in questions get.
> 4. For smaller objects, the pooling overheads become on par with the size of the objects themselves. The calculation for total memory footprint can push the scale in either direction.
> 5. For some _awkward_ classes like DirectByteBuffers that have separate cleanup schedule, unbounded allocation is a recipe for a meltdown.
>
> So answer is somewhat along the lines of: Pooling common (small) objects? Nah, too much hassle for too little gain. Pooling large buffers? Yes, that is a common perf optimization. Pooling large buffers with special lifecycle? YES, do not even _think_ about _not_ doing the pooling. For everything in between the answer is somewhere in between.

Here, "special lifecycle" refers to things like finalizers, Cleaners, weak references, etc.; nothing that would apply to a simple byte buffer.

Another interesting point that came up is that if you use heap (non-direct) byte buffers, and if the pool doesn't hold on to byte buffer references while they are leased out, then there is no risk of a memory leak: returning the buffer to the pool is purely an optimization. Since `HEAP` and `DIRECT` have near-identical performance, maybe we should just hardcode a pooled heap buffer allocator into key hotspots.

ok2c:
@rschmitt Thank you so much for such an informative summary. Please convey my gratitude to Aleksey.

One thing bugs me is how big is big? How big should be byte buffers to justify pooling? If it is a couple of MB, then memory pooling may be useful in our case.

> Here, "special lifecycle" refers to things like finalizers, Cleaners, weak references, etc.; nothing that would apply to a simple byte buffer.

I think we have objects with "special lifecycle" in the HttpClient Caching module only but they are backed by files and not byte buffers. There is nothing else I can think of.

However, I imagine the classic on async facade may actually qualify as a potential beneficiary of the pooled memory allocator, so I am leaning towards approving this change-set and letting @arturobernalg proceed with further experiments.

What do you think?

arturobernalg:
@olegk @rschmitt How should I proceed here?

ok2c:
@arturobernalg I agree with @rschmitt here. There is no point adding this complex feature just in case. We need to have a strong case where it actually improves performance. I think it could really help with classic over async facade, but this needs to be carefully measured and benchmarked. I would propose you start by profiling the classic over async facade classes.

rschmitt:
Why the classic-over-async facade?

ok2c:
> Why the classic-over-async facade?

@rschmitt It was my suggestion. See my comment above. If you have a better idea where this feature can be applied please do suggest.

ok2c:
@arturobernalg This is what I think. Micro-benchmarks are nice and they can be useful, but ultimately what matters if a performance optimization feature makes a real and verifiable impact on request execution.  This is the benchmark I have been using while working on HttpClient 5.x performance optimization. I can still be useful to collect total execution time / request per seconds metrics of different agents or different message types. It is not very scientific but if it is usually good enough for better / worse performance comparison. 

If you can find a scenario where pooled ByteBuffer allocator makes a noticeable and verifiable impact, say, of 5% improvement, then it is useful. If not, I am sorry to say I do not think it should be added to the project code.

I think there is a chance of making classic over async with very large messages to perform better with this pooled ByteBuffer allocator but I would like to see hard evidence of that.    

[1] https://github.com/ok2c/httpclient-benchmark

arturobernalg:
Hi @ok2c ,
                                                                                                                                                                                                                    
  I've been benchmarking different ByteBufferAllocator strategies for the classic-over-async facade with 1MB responses at concurrency 50.
                                                                                                                                                                                                                    
  Key finding: A simple ThreadLocalByteBufferAllocator that caches the single largest released buffer per thread significantly outperforms both the baseline and the bucketed PooledByteBufferAllocator.            
                                                                                                                                                                                                                    
  JMH microbenchmark results (HEAP, ops/ms, higher is better):                                                                                                                                                      
                                                            
  ┌─────────────┬───────────────────┬────────┬─────────────┐                                                                                                                                                        
  │ Buffer Size │ Simple (baseline) │ Pooled │ ThreadLocal │
  ├─────────────┼───────────────────┼────────┼─────────────┤                                                                                                                                                        
  │ 1KB         │ 174               │ 260    │ 367         │
  ├─────────────┼───────────────────┼────────┼─────────────┤
  │ 8KB         │ 22                │ 88     │ 90          │                                                                                                                                                        
  ├─────────────┼───────────────────┼────────┼─────────────┤                                                                                                                                                        
  │ 64KB        │ 2.6               │ 11.5   │ 11.6        │                                                                                                                                                        
  └─────────────┴───────────────────┴────────┴─────────────┘                                                                                                                                                        
                                                            
  Both Pooled and ThreadLocal show zero GC allocations on the hot path (gc.count ≈ 0).                                                                                                                              
                                                            
  End-to-end benchmark (50K GET requests, 1MB response, c=50, 4 rounds with alternating order):                                                                                                                     
                                                            
  ┌─────────────┬─────────────┬─────────────┐                                                                                                                                                                       
  │    Agent    │ Avg req/sec │ vs Baseline │               
  ├─────────────┼─────────────┼─────────────┤                                                                                                                                                                       
  │ Baseline    │ 1422        │ —           │               
  ├─────────────┼─────────────┼─────────────┤
  │ Pooled      │ 1442        │ +1.4%       │
  ├─────────────┼─────────────┼─────────────┤                                                                                                                                                                       
  │ ThreadLocal │ 1567        │ +10.2%      │
  └─────────────┴─────────────┴─────────────┘                                                                                                                                                                       
                                                            
  Why ThreadLocal wins: The IO reactor uses a small fixed thread pool. After the first request each thread's buffer stabilizes at the final size. Subsequent requests get a direct cache hit — zero expandCapacity()
   chains, zero copies, zero GC. The PooledByteBufferAllocator achieves the same reuse but pays for bucket lookups and CAS operations that the ThreadLocal approach avoids entirely.
                                                                                                                                                                                                                    
  The implementation is ~30 lines: one ThreadLocal<ByteBuffer>, keep-largest-on-release policy, no locks, no contention.                                                                                            
   
  I'm going to prepare the PR with the ThreadLocalByteBufferAllocator and the full benchmark results.

ok2c:
@arturobernalg `ThreadLocal`s usually make me uncomfortable, `ThreadLocal`s that can hold MBs of data make me very, very uncomfortable. It is nice it shows improvement in the benchmark results but it needs to be absolutely leak-proof for real life scenarios.

rschmitt:
> `ThreadLocal`s that can hold MBs of data make me very, very uncomfortable

Two things worth keeping in mind here:

1. Virtual threads make it economical to have on the order of 10,000 threads.
2. Java is currently in the hot seat due to the DRAM shortage. A basic aspect of the JVM's design is to use the largest heap possible, because that results in more efficient garbage collection, but just over the last few months this is starting to become expensive. Reducing ephemeral garbage is still a good idea as far as it goes, but now would be a particularly bad time to increase memory _overhead_, i.e. the baseline amount of tenured heap memory we need to run.

arturobernalg:
Hi @ok2c ,

  I've spent considerable time benchmarking the classic-over-async facade trying to reach the 5% improvement threshold. Here's a summary of what I found.

  Benchmark setup: 50,000 requests, concurrency 200, 1MB response bodies, HTTP/1.1, 6 rounds per configuration, A/B comparison in the same session where possible.

  What I tried:

  1. `PooledByteBufferAllocator` wired through `ClassicToAsyncAdaptor` → `SharedInputBuffer`: Result within noise (±2%).   during buffer expansion (2KB→4KB→...→2MB), only intermediate
  buffers are recycled. The final ~2MB buffer cannot be safely returned to the pool because `releaseResources()` races with async framework callbacks (use-after-free). So every request still allocates a fresh final
   buffer.
  2. Content-Length pre-sizing (read Content-Length header and pre-allocate `SharedInputBuffer` to the right size): Actually 16% slower. With 200 concurrent connections, pre-allocating 200×1MB buffers up front
  overwhelms the system and disrupts the flow control dynamics (capacity channel reports 1MB available instead of 2KB).
  3. `SharedInputBuffer` micro-optimizations (signalAll() → signal(), AtomicInteger → plain int for capacity increment since it's always accessed under the lock): Correct improvements semantically, but not
  measurable — within noise.

  Conclusion: At concurrency 200 with 1MB responses, the system is transferring ~2.5GB/sec of content. The bottleneck is memory bandwidth and network I/O, not buffer allocation or lock contention in
  `SharedInputBuffer`. The buffer management overhead is a rounding error compared to the cost of moving data through the network stack.

  The `PooledByteBufferAllocator` itself performs well in isolation (JMH: 471 ops/ms vs 194 for `SimpleByteBufferAllocator` at 1KB HEAP), but the classic-over-async facade's one-way buffer growth pattern prevents the pool from being effective. The allocator would likely show better results in code paths where buffers are allocated and released at the same size repeatedly, like async framework internals or H2 frame
  handling.

  I'll continue testing other code paths this week to find where the allocator can demonstrate measurable improvement.

arturobernalg:
Following up on the benchmarking: I couldn't get the pooled allocator to show a verifiable improvement — end-to-end it's within noise (data below), which matches the concerns raised here. The profiling did surface the actual bottleneck though: the 8 KB default `Http1Config` buffer throttles large-message 
throughput. Raising it to 32 KB is ~+41% on 1 MB bodies with all tests green. I'll close this and open a small PR for that instead. Thanks for the pushback — it pointed to the right place.

PR #598, opened 2026-01-17
Add CompletableFuture and CompletionStage accessors to ReactiveResponseConsumer

arturobernalg:
Expose response availability, exchange completion, and terminal failure as CompletionStage backed by CompletableFuture.

### DIFF
--- a/httpcore5-reactive/src/main/java/org/apache/hc/core5/reactive/ReactiveResponseConsumer.java
+++ b/httpcore5-reactive/src/main/java/org/apache/hc/core5/reactive/ReactiveResponseConsumer.java
@@ -32,6 +32,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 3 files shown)


httpcore5-reactive/src/main/java/org/apache/hc/core5/reactive/ReactiveResponseConsumer.java
```diff
@@ -144,10 +165,10 @@ public List<Header> getTrailers() {
 
     @Override
     public void consumeResponse(
-        final HttpResponse response,
-        final EntityDetails entityDetails,
-        final HttpContext httpContext,
-        final FutureCallback<Void> resultCallback
+            final HttpResponse response,
```
ok2c: @arturobernalg Formatting noise

rschmitt:
I don't know what to make of all this. What problem are you trying to solve?

ok2c:
@rschmitt I imagine this could be useful to users but this is your call

PR #654, opened 2026-03-18
Implement two-phase graceful HTTP/2 shutdown

arturobernalg:
This change implements two-phase graceful shutdown for HTTP/2 connections.
The multiplexer sends an initial GOAWAY, waits one RTT using PING, then sends the final GOAWAY with the last processed stream identifier while continuing to handle in-flight peer streams and peer GOAWAY boundaries correctly.

### DIFF
--- a/httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/AbstractH2StreamMultiplexer.java
+++ b/httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/AbstractH2StreamMultiplexer.java
@@ -106,7 +106,7 @@ abstract class AbstractH2StreamMultiplexer implements Identifiable, HttpConnecti
 
     private static final long CONNECTION_WINDOW_LOW_MARK = 10 * 1024 * 1024;
 
-    enum ConnectionHandshake { READY, ACTIVE, GRACEFUL_SHUTDOWN, SHUTDOWN }
+    enum ConnectionHandshake { READY, ACTIVE, DRAINING, GRACEFUL_SHUTDOWN, SHUTDOWN }
     enum SettingsHandshake { READY, TRANSMITTED, ACKED }
 
     private final ProtocolIOSession ioSession;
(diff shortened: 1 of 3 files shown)


ok2c:
@arturobernalg Why is this needed? What does this improve excactly?

arturobernalg:
> Why is this needed? What does this improve excactly?

@ok2c 
RFC 9113 recommends two-phase graceful shutdown for HTTP/2: initial GOAWAY with 2^31-1, wait one RTT, then final GOAWAY with the actual last processed stream id. This improves correctness of connection draining by avoiding shutdown races around in-flight stream creation and by preserving an accurate final GOAWAY boundary.

https://www.rfc-editor.org/rfc/rfc9113.html#section-6.8

“A server that is attempting to gracefully shut down a connection SHOULD send an initial GOAWAY frame with the last stream identifier set to 2^31-1 and a NO_ERROR code.”

“After allowing time for any in-flight stream creation (at least one round-trip time), the server MAY send another GOAWAY frame with an updated last stream identifier.”

ok2c:
> A server that is attempting 

@arturobernalg Very well. A server. And what about clients? Does this make any sense for client connections?

arturobernalg:
> > A server that is attempting
> 
> @arturobernalg Very well. A server. And what about clients? Does this make any sense for client connections?

Good point.For client connections, sending GOAWAY on graceful close still makes sense, but the two-phase drain logic appears to have much less practical value except perhaps for server push.

ok2c:
@arturobernalg What I was trying to say this feature would make more sense for the server side and so much for the client side, where it would just slow down connection termination. It may still be worth looking at but for the server side only.

arturobernalg:
Hi @ok2c 

I updated the patch to restrict the two-phase graceful shutdown logic to server-side connections only.
The server now sends an initial GOAWAY, waits one RTT using PING, and then sends the final GOAWAY with the last processed stream id. Client-side graceful shutdown remains a single-GOAWAY path. Please take a look https://github.com/apache/httpcomponents-core/pull/661

PR #657, opened 2026-03-30
Fix H2 stream timeout handling on inbound frames

arturobernalg:
This patch fixes H2 stream timeout handling for delayed inbound frames.

Expired streams are now reset before inbound DATA, HEADERS, CONTINUATION, or PUSH_PROMISE frames can refresh stream activity. Regression tests have been added for the affected code paths.

### DIFF
--- a/httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/AbstractH2StreamMultiplexer.java
+++ b/httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/AbstractH2StreamMultiplexer.java
@@ -782,12 +782,17 @@ abstract class AbstractH2StreamMultiplexer implements Identifiable, HttpConnecti
         if (frameType == null) {
             return;
         }
+        final long nowNanos = System.nanoTime();
         switch (frameType) {
             case DATA: {
                 if (streamId == 0) {
                     throw new H2ConnectionException(H2Error.PROTOCOL_ERROR, "Illegal stream id: " + streamId);
                 }
                 final H2Stream stream = streams.lookupValid(streamId);
+                if (resetIfExpired(stream, nowNanos)) {
+                    requestSessionOutput();
+                    break;
+                }
                 try {
                     consumeDataFrame(frame, stream);
                 } catch (final H2StreamResetException ex) {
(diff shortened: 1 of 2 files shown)


httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/AbstractH2StreamMultiplexer.java
```diff
@@ -782,12 +782,17 @@ private void consumeFrame(final RawFrame frame) throws HttpException, IOExceptio
         if (frameType == null) {
             return;
         }
+        final long nowNanos = System.nanoTime();
```
ok2c, line 785: @arturobernalg Why are we using `#nanoTime` here when all we need is a second precision? How is it better? I suppose it is certinaly not cheaper then `#currentTimeMillis`

arturobernalg: @ok2c 
`nanoTime()` is not better because of finer precision. It is better because it is monotonic.
`currentTimeMillis()` is wall-clock time since the epoch, while `nanoTime()` is intended for elapsed-time measurement and uses a monotonic timer; its actual resolution still depends on OS and hardware.

https://docs.oracle.com/cd/E15289_01/JRDTG/slow_start.htm

ok2c: > @ok2c `nanoTime()` is not better because of finer precision. It is better because it is monotonic. `currentTimeMillis()` is wall-clock time since the epoch, while `nanoTime()` is intended for elapsed-time measurement and uses a monotonic timer; its actual resolution still depends on OS and hardware.

@arturobernalg OK. Fair enough. I would prefer this change be done consistency across the entire code base. Not a bit here and a bit there.

httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/AbstractH2StreamMultiplexer.java
```diff
@@ -830,6 +835,10 @@ private void consumeFrame(final RawFrame frame) throws HttpException, IOExceptio
                     stream.activate();
                     streams.resetIfExceedsMaxConcurrentLimit(stream, localConfig.getMaxConcurrentStreams());
                 }
+                if (resetIfExpired(stream, nowNanos)) {
+                    requestSessionOutput();
```
ok2c: @arturobernalg This is unnecessary. Any outgoing frame automatically triggers output interest automatically.

ok2c (approved):
@arturobernalg Please remove unnecessary code. Looks good otherwise.

PR #661, opened 2026-04-02
Restrict two-phase graceful HTTP/2 shutdown to server-side connections

arturobernalg:
Send an initial GOAWAY, wait one RTT using PING, then send the final GOAWAY with the last processed stream id. Keep client-side graceful shutdown as a single-GOAWAY path.


ok2c:
@arturobernalg Please move this logic to `ServerH2StreamMultiplexer`. If it cannot be cleanly decoupled and moved, drop the whole thing. It is not worth it.

arturobernalg:
> @arturobernalg Please move this logic to `ServerH2StreamMultiplexer`. If it cannot be cleanly decoupled and moved, drop the whole thing. It is not worth it.

@ok2c  please do another pass

ok2c:
@arturobernalg No, we are not doing this. Drop this change-set. Please focus on something useful.

PR #662, opened 2026-04-03
Configurable HTTP/2 PING ACK timeout

arturobernalg:
This change makes the HTTP/2 PING ACK timeout configurable through H2Config instead of using a hard-coded 5 second value in the multiplexer idle validation path


ok2c:
@arturobernalg Why? What is the use case?

arturobernalg:
> @arturobernalg Why? What is the use case?

@ok2c 
he use case is high-latency or intermittently stalled links where the fixed timeout can cause false connection shutdowns during validation.

ok2c:
@arturobernalg I kind of think that 5 seconds should be plentiful to respond to a lone ping. If you still want to make this configurable please do not abuse `H2Config` object that represents H2 protocol parameters. Use constructor parameter, similar to `validateAfterInactivity`

arturobernalg:
> @arturobernalg I kind of think that 5 seconds should be plentiful to respond to a lone ping. If you still want to make this configurable please do not abuse `H2Config` object that represents H2 protocol parameters. Use constructor parameter, similar to `validateAfterInactivity`

Done

ok2c (approved):

PR #511, opened 2025-01-03
HTTPCORE-774: fixed a race condition caused by concurrent update of the connection input window to the max value

ok2c:
@arturobernalg Please review if you happen to have capacilty

PR #672, opened 2026-06-15
HTTPCORE-774: apply missing connection input window fix to 5.4.x

osipxd:
This cherry-picks d31d0000084d1ffdd37a9596b53e2aa7779451bc (#511) onto 5.4.x. The companion HTTPCORE-774 window management commit is already present on 5.4.x (2cb5fb73faf85cbc4efa380d8ee9c442b94831f3), but this one-line connection input window fix appears to be missing.

Probably should also be cherry-picked onto `main`.

### DIFF
--- a/httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/AbstractH2StreamMultiplexer.java
+++ b/httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/AbstractH2StreamMultiplexer.java
@@ -1492,7 +1492,6 @@ abstract class AbstractH2StreamMultiplexer implements Identifiable, HttpConnecti
 
         @Override
         public void update(final int increment) throws IOException {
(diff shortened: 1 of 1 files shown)


ok2c:
@osipxd  Hmm. Bizarre. I will investigate

ok2c:
@osipxd This is so odd. I intentionally split the fix into change-sets: the actual fix and some code improvements around it. I imagine I wanted the actual fix to go into the stable branch, which was 5.3.x at that time, and both change-sets into the current dev branch which was master (5.4.x). Both change-sets however are present in 5.3

```
$ git log rel/v5.3.6 --oneline | grep HTTPCORE-774
2a83df38b HTTPCORE-774: update window management improvements
d31d00000 HTTPCORE-774: fixed a race condition caused by concurrent update of the connection input window to the max value.
```
and the 5.4.x has improvements only and no actual fix.

I suppose I am getting old.

ok2c:
@osipxd Also cherry-picked to master

PR #673, opened 2026-06-16
Remove @Experimental for QUERY method and implement conformance to RFC 10008

desiderantes:
###  Removing Experimental Markings

Since the QUERY method draft has been promoted to a proper [RFC]( https://www.rfc-editor.org/rfc/rfc10008.html), this change removes the @Experimental tag from   QUERY method.

### RFC 10008 Conformance

  According to Section 2 of RFC 10008 https://www.rfc-editor.org/rfc/rfc10008.html:

  │ Servers MUST fail the request if the Content-Type request field is missing or is inconsistent with the request content.

  I've added code to enforce this constraint.
  
  This is follow up of #499

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,8 @@ Protocol conformance
 - [RFC 1945](https://datatracker.ietf.org/doc/html/rfc1945) - Hypertext Transfer Protocol -- HTTP/1.0
 - [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986) - Uniform Resource Identifier (URI): Generic Syntax
 - [RFC 9218](https://datatracker.ietf.org/doc/html/rfc9218) - Extensible Prioritization Scheme for HTTP
+- [RFC 10008](https://datatracker.ietf.org/doc/html/rfc10008) - The HTTP QUERY Method
+
 
 Licensing
 ---------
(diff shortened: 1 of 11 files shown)


ok2c:
@desiderantes I will do my best to get around to reviewing this change-set and including it in the upcoming 5.5-beta2 release. Please note though we usually do not upgrade to RFCs piecemeal. We need to be sure that at the very least all hard requirements are met

Please also add RFC 10008 to the list of RFCs the library conforms to in README.md

desiderantes:
> @desiderantes I will do my best to get around to reviewing this change-set and including it in the upcoming 5.5-beta2 release. Please note though we usually do not upgrade to RFCs piecemeal. We need to be sure that at the very least all hard requirements are met

Hi, in this case, the prior change implemented basic QUERY method support, while the RFC was in draft status. After it graduated from that status, there were now missing parts for conformance, which I add with this PR.

> Please also add RFC 10008 to the list of RFCs the library conforms to in README.md

Done

ok2c:
> Hi, in this case, the prior change implemented basic QUERY method support, while the RFC was in draft status. After it graduated from that status, there were now missing parts for conformance, which I add with this PR.

@desiderantes What I am trying to say we need to be sure the protocol code fully conforms to RFC 10008 in terms of hard requirements before we can remove `experimental` annotation. If your change-set already covers all those requirements then all is well.

desiderantes:
> > Hi, in this case, the prior change implemented basic QUERY method support, while the RFC was in draft status. After it graduated from that status, there were now missing parts for conformance, which I add with this PR.
> 
> @desiderantes What I am trying to say we need to be sure the protocol code fully conforms to RFC 10008 in terms of hard requirements before we can remove `experimental` annotation. If your change-set already covers all those requirements then all is well.

Ah, got you, no problem. Yes, the required changes in this project were minimal from my point of view. There is another PR for the client itself that handles the changes required there, and I referenced this PR from that one.

PR #674, opened 2026-06-17
SSLIOSession: Fix regression in backpressure handling

rschmitt:
A regression was introduced by PR #513, submitted as a fix for HTTPCORE-775, which claimed that `SSLIOSession::write` was failing to "handle" `SSLEngineResult#BUFFER_OVERFLOW`. The issue was encountered by Apache CXF, who provided a test case demonstrating the problem:

https://github.com/apache/cxf/pull/2214/changes

I investigated this reproducer and found that CXF was actually hanging due to not signaling the availability of more data:

https://github.com/apache/cxf/blob/f1b2c37bd9f3606cf7ca2d5d39cc1168e94fd9e4/rt/transports/http-hc5/src/main/java/org/apache/cxf/transport/http/asyncclient/hc5/CXFHttpAsyncRequestProducer.java#L97

when `SSLIOSession#write` signaled backpressure here:

https://github.com/apache/cxf/blob/f1b2c37bd9f3606cf7ca2d5d39cc1168e94fd9e4/rt/transports/http-hc5/src/main/java/org/apache/cxf/transport/http/asyncclient/hc5/CXFHttpAsyncRequestProducer.java#L75

The return value of `buf.produceContent` (which is ultimately `SSLEngineResult#bytesConsumed` -- 0, in this case) is simply being discarded. Changing the `CXFHttpAsyncRequestProducer#available` method as follows causes the problematic test cases to succeed instantly:

```
    @Override
    public int available() {
        if (buffer != null && buffer.hasRemaining()) {
            return buffer.remaining();
        } else if (buf != null && buf.length() > 0) {
            return buf.length();
        }
        return 0;
    }
```

Additionally, that PR appears to have introduced dysfunctional backpressure handling into `SSLIOSession` by simply expanding the buffer as much as necessary to hold the encrypted `src` buffer. This in turn has caused a significant regression in heap usage and latency, hence the need to revert. I've retained the test changes, since they did introduce more coverage for TLSv1.3.

### DIFF
--- a/httpcore5-testing/src/test/java/org/apache/hc/core5/testing/nio/Http1IntegrationTest.java
+++ b/httpcore5-testing/src/test/java/org/apache/hc/core5/testing/nio/Http1IntegrationTest.java
@@ -750,6 +750,9 @@ abstract class Http1IntegrationTest extends HttpIntegrationTest {
 
         @Override
         public int write(final ByteBuffer src) throws IOException {
+            if (!channel().isOpen()) {
+                return 0;
+            }
             final int chunk;
             if (!done) {
                 lineBuffer.clear();
--- a/httpcore5/src/main/java/org/apache/hc/core5/reactor/ssl/SSLIOSession.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/reactor/ssl/SSLIOSession.java
@@ -93,7 +93,6 @@ public class SSLIOSession implements IOSession {
     private final AtomicInteger outboundClosedCount;
     private final AtomicReference<TLSHandShakeState> handshakeStateRef;
     private final IOEventHandler internalEventHandler;
-    private final int packetBufferSize;
 
     private int appEventMask;
 
@@ -180,9 +179,9 @@ public class SSLIOSession implements IOSession {
 
         final SSLSession sslSession = this.sslEngine.getSession();
         // Allocate buffers for network (encrypted) data
-        this.packetBufferSize = sslSession.getPacketBufferSize();
-        this.inEncrypted = SSLManagedBuffer.create(sslBufferMode, packetBufferSize);
-        this.outEncrypted = SSLManagedBuffer.create(sslBufferMode, packetBufferSize);
+        final int netBufferSize = sslSession.getPacketBufferSize();
+        this.inEncrypted = SSLManagedBuffer.create(sslBufferMode, netBufferSize);
+        this.outEncrypted = SSLManagedBuffer.create(sslBufferMode, netBufferSize);
 
         // Allocate buffers for application (unencrypted) data
         final int appBufferSize = sslSession.getApplicationBufferSize();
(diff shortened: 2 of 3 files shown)


ok2c:
@rschmitt Thank you for the thorough analysis. There is a single test failing on MacOS only. Ideally this should be fixed. Otherwise, I have to agree I have made a mistake reviewing the proposed change-set

rschmitt:
@ok2c The `Http1IntegrationTest.testTruncatedChunk` flakiness is 100% related to this PR somehow. I'll get to the bottom of it.

reta:
> @rschmitt Thank you for the thorough analysis. There is a single test failing on MacOS only. Ideally this should be fixed. Otherwise, I have to agree I have made a mistake reviewing the proposed change-set

@ok2c (not you to blame) sorry about that folks, I think what derailed me from suspecting CXF is that the issue manifested suddenly with just bumping the HttpCore5 version, and only with TLSv1.3 (coming from JDK defaults).

rschmitt:
I found something interesting while debugging. `SSLEngine#wrap` will consume zero bytes and report overflow unless the `dst` buffer has at least `sslSession.getPacketBufferSize()` bytes available. You can see this by setting

```java
this.outEncrypted = SSLManagedBuffer.create(sslBufferMode, netBufferSize - 1);
```

which will cause TLS requests to hang. This means that we can only perform one call to `SSLIOSession#write` per reactor loop, even if the writes are small enough in aggregate to fit in the output buffer. Evidently, `wrap` [always produces an Application Data packet](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/javax/net/ssl/SSLEngine.html), therefore the output buffer must first have enough space for the largest possible packet:

> The SSLEngine produces/consumes complete SSL/TLS/DTLS packets only, and does not store application data internally between calls to wrap()/unwrap(). Thus input and output ByteBuffers must be sized appropriately to hold the maximum record that can be produced. Calls to [SSLSession.getPacketBufferSize()](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/javax/net/ssl/SSLSession.html#getPacketBufferSize()) and [SSLSession.getApplicationBufferSize()](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/javax/net/ssl/SSLSession.html#getApplicationBufferSize()) should be used to determine the appropriate buffer sizes. The size of the outbound application data buffer generally does not matter. If buffer conditions do not allow for the proper consumption/production of data, the application must determine (via [SSLEngineResult](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/javax/net/ssl/SSLEngineResult.html)) and correct the problem, and then try the call again.

This introduces a trade-off between latency and heap allocation. We might consider whether to use a _slightly_ larger `outEncrypted` buffer so that we can get the first Application Data packet sent off immediately even if a handshake packet like ChangeCipherSpec is still sitting in the buffer; probably just a few dozen bytes would make a difference. The broader question is whether or how to support a _much_ larger output buffer so that we can batch encrypt many packets in a single event loop iteration.

rschmitt:
The test failure was caused by `BrokenChunkEncoder#write` failing to handle async flushes correctly. The backpressure handling allows a state where the channel is marked closed, but data is still being flushed asynchronously. The call into `BrokenChunkEncoder#write` while in this state was triggering a `ChannelClosedException`, which got translated into a TCP RST sent from the server to the client, which then never saw the malformed chunk, the partial chunk data, or the expected `MalformedChunkCodingException`.

ok2c:
> This introduces a trade-off between latency and heap allocation. We might consider whether to use a _slightly_ larger `outEncrypted` buffer so that we can get the first Application Data packet sent off immediately even if a handshake packet like ChangeCipherSpec is still sitting in the buffer; probably just a few dozen bytes would make a difference. The broader question is whether or how to support a _much_ larger output buffer so that we can batch encrypt many packets in a single event loop iteration.

@rschmitt This decision could be delegated to a strategy interface or a new method in `SSLSessionInitializer`.

ok2c (approved):
@rschmitt Please also cheery-pick to `5.4.x`

ok2c:
Cherry-picked to `5.4.x`

PR #675, opened 2026-06-20
HTTPCORE-796: abort redirected requests during the expect-continue handshake the same way as errors

ok2c:


arturobernalg (approved):
Looks good to me.

rschmitt (approved):

PR #683, opened 2026-07-23
H2 stream handlers to make sure content-length header value equals the sum of data payload

ok2c:
H2 stream handlers to make sure content-length header value equals the sum of data payload.

Conformance to RFC 9113, section 8.1.1

### DIFF
--- a/httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/ClientH2StreamHandler.java
+++ b/httpcore5-h2/src/main/java/org/apache/hc/core5/http2/impl/nio/ClientH2StreamHandler.java
@@ -41,10 +41,12 @@ import org.apache.hc.core5.http.HttpRequest;
 import org.apache.hc.core5.http.HttpResponse;
 import org.apache.hc.core5.http.HttpStatus;
 import org.apache.hc.core5.http.HttpVersion;
+import org.apache.hc.core5.http.Method;
 import org.apache.hc.core5.http.ProtocolException;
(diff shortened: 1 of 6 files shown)


arturobernalg:
@ok2c 
Shouldn't the content length also be validated when trailer HEADERS terminate the stream? [RFC 9113 §8.1](https://www.rfc-editor.org/rfc/rfc9113.html#section-8.1) states that trailer fields are carried in a field block that terminates the stream, while [§8.1.1](https://www.rfc-editor.org/rfc/rfc9113.html#section-8.1.1) requires Content-Length to equal the sum of the DATA frame payload lengths.

ok2c:
> Shouldn't the content length also be validated when trailer HEADERS terminate the stream? [RFC 9113 §8.1](https://www.rfc-editor.org/rfc/rfc9113.html#section-8.1) states that trailer fields are carried in a field block that terminates the stream, while [§8.1.1](https://www.rfc-editor.org/rfc/rfc9113.html#section-8.1.1) requires Content-Length to equal the sum of the DATA frame payload lengths.

@arturobernalg I honestly do not know. Do you think it should? If you do, I will update the change-set. As far as I can tell, in practical terms this check matters for message head only due to a possible H2 to HTTP/1.1 translation by an intermediary.

arturobernalg:
> I honestly do not know. Do you think it should? If you do, I will update the change-set. As far as I can tell, in practical terms this check matters for message head only due to a possible H2 to HTTP/1.1 translation by an intermediary.

@ok2c 
I do not consider it a critical blocker if it would rather handle it separately.

arturobernalg (approved):
LGTM

PR #689, opened 2026-08-15
Add a model and parser for the RFC 9209 Proxy-Status response field

arturobernalg:
Proxy-Status is a Structured Fields (RFC 8941) List whose members carry an intermediary identity and a set of parameters. ProxyStatus is an immutable holder for a single member, ProxyStatusError enumerates the standardized error tokens, and ProxyStatusSupport parses the field value. Parameter values keep their structured-field types, the standardized and error-specific parameter types are validated, received-status is checked as an HTTP status code, and byte-sequence values are defensively copied. The code only decodes the field; it neither interprets nor acts on the reported values.

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/http/support/ProxyStatusSupport.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/http/support/ProxyStatusSupport.java
@@ -1,0 +1,383 @@
+/*
+ * ====================================================================
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ * ====================================================================
+ *
+ * This software consists of voluntary contributions made by many
+ * individuals on behalf of the Apache Software Foundation.  For more
+ * information on the Apache Software Foundation, please see
+ * <http://www.apache.org/>.
+ *
+ */
+
+package org.apache.hc.core5.http.support;
+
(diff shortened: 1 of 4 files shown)


httpcore5/src/main/java/org/apache/hc/core5/http/support/ProxyStatusSupport.java
```diff
@@ -0,0 +1,383 @@
...
+    /**
+     * Parses a {@code Proxy-Status} field value into its ordered list of members.
+     *
+     * @param value the field value; must not be {@code null}.
+     * @return the parsed members, empty when the value contains no members.
+     * @throws ParseException if the value does not conform to RFC 9209 / RFC 8941.
+     */
+    public static List<ProxyStatus> parse(final CharSequence value) throws ParseException {
+        Args.notNull(value, "Proxy-Status value");
+        final ParserCursor cursor = new ParserCursor(0, value.length());
```
ok2c, line 74: @arturobernalg Keep it flexible and pass `ParserCursor` as a parameter. That would enable the parse message to take its input from directly `FormattedHeader` without having to make an intermediate String out of it. 

See `MessageSupport#parserHeaderValue`

ok2c:
@arturobernalg Overall, what is your plan for this code? Is this something that you need at work or for a private project? I do not see this code applicable anywhere in neither core or client at  the moment. What is the plan?

arturobernalg:
> Overall, what is your plan for this code? Is this something that you need at work or for a private project? I do not see this code applicable anywhere in neither core or client at the moment. What is the plan?

It is not for a specific work or private project. My intention was to add RFC 9209 support incrementally. I started with Core because the field syntax and model are generic HTTP semantics rather than client policy.

ok2c:
> > Overall, what is your plan for this code? Is this something that you need at work or for a private project? I do not see this code applicable anywhere in neither core or client at the moment. What is the plan?
> 
> It is not for a specific work or private project. My intention was to add RFC 9209 support incrementally. I started with Core because the field syntax and model are generic HTTP semantics rather than client policy.

@arturobernalg Your plan is to provide it out of client? If so, this code should go to client. It can be moved to core once considered stable and there is an actual need for it.

arturobernalg:
> > > Overall, what is your plan for this code? Is this something that you need at work or for a private project? I do not see this code applicable anywhere in neither core or client at the moment. What is the plan?
> > 
> > 
> > It is not for a specific work or private project. My intention was to add RFC 9209 support incrementally. I started with Core because the field syntax and model are generic HTTP semantics rather than client policy.
> 
> @arturobernalg Your plan is to provide it out of client? If so, this code should go to client. It can be moved to core once considered stable and there is an actual need for it.

@ok2c . That makes sense — I’ll move the implementation to Client first and keep it there until there is a demonstrated need to expose it from Core.

PR #690, opened 2026-08-22
HTTPCORE-798: Make socket channel creation customizable

arturobernalg:
Add a socket channel factory extension point to the connecting I/O reactor.

### DIFF
--- a/httpcore5/src/main/java/org/apache/hc/core5/reactor/SingleCoreIOReactor.java
+++ b/httpcore5/src/main/java/org/apache/hc/core5/reactor/SingleCoreIOReactor.java
@@ -54,6 +54,7 @@ import org.apache.hc.core5.function.Callback;
 import org.apache.hc.core5.function.Decorator;
 import org.apache.hc.core5.io.CloseMode;
 import org.apache.hc.core5.io.Closer;
+import org.apache.hc.core5.io.IOFunction;
 import org.apache.hc.core5.net.NamedEndpoint;
 import org.apache.hc.core5.util.Args;
 import org.apache.hc.core5.util.ReflectionUtils;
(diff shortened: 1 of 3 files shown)


httpcore5/src/main/java/org/apache/hc/core5/reactor/SocketChannelFactory.java
```diff
@@ -0,0 +1,56 @@
...
+ * <p>
+ * The channel returned by this factory must be open and unconnected.
+ * The I/O reactor remains responsible for channel configuration, local
+ * binding, connection establishment, selector registration and lifecycle.
+ * </p>
+ *
+ * @since 5.5
+ */
+@FunctionalInterface
+public interface SocketChannelFactory {
```
ok2c: @arturobernalg Can't we reuse already existing `IOFunction` interface here instead of this one?

arturobernalg: > @arturobernalg Can't we reuse already existing `IOFunction` interface here instead of this one?

@ok2c Good point,  I’ll reuse it and remove `SocketChannelFactory`.

httpcore5/src/main/java/org/apache/hc/core5/reactor/SingleCoreIOReactor.java
```diff
@@ -349,8 +365,8 @@ private void processPendingConnectionRequests() {
             if (!sessionRequest.isCancelled()) {
                 final SocketChannel socketChannel;
                 try {
-                    socketChannel = openSocketFor(sessionRequest.remoteAddress);
-                } catch (final IOException ex) {
+                    socketChannel = Args.notNull(socketChannelFactory.apply(sessionRequest.remoteAddress), "Socket channel");
```
ok2c: @arturobernalg This looks odd. Why `Args#notNull` here? It is not an argument and it is never `null`

ok2c (approved):

PR #696, opened 2026-09-11
Fix URIBuilder encoding policy documentation and null reset

prafgup:
URIBuilder defaults to ALL_RESERVED, but setEncodingPolicy's Javadoc incorrectly identifies RFC_3986 as the default. Passing null also selects RFC_3986 behavior instead of restoring the documented default.

This change corrects the Javadoc and makes null restore ALL_RESERVED.

Incorrect doc - 
- https://github.com/apache/httpcomponents-core/blob/4021b98e8e942ccbce8cdf54ec910302821fa740/httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java#L206-L216

Current Default - 
- https://github.com/apache/httpcomponents-core/blob/4021b98e8e942ccbce8cdf54ec910302821fa740/httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java#L91

Validation: ./mvnw -pl httpcore5 verify passed. All 1,380 tests passed, with no failures, errors, or skips.


ok2c:
Cherry-picked to `5.4.x`

PR #698, opened 2026-09-17
HTTPCORE-799: Reset async entity producer state after interrupted writes

mkurz:
This fixes [HTTPCORE-799](https://issues.apache.org/jira/browse/HTTPCORE-799).

This was found while testing an HTTP/2 authentication retry in which a server returned a `401` response and then reset the stream with `RST_STREAM(NO_ERROR)`. The request producer was released mid-write and then reused for the retry.

Repeatable async entity producers could retain state from an interrupted write when `releaseResources()` was called. Although the source was reset for a replay, internal staging buffers could still contain the unwritten tail, and a digesting decorator could still contain the bytes accepted before the interruption.

The affected producers now reset their per-production state when resources are released:

- `FileEntityProducer` and `PathEntityProducer` clear their file buffers.
- `AbstractCharAsyncEntityProducer` clears its encoded-byte buffer, fixing `StringAsyncEntityProducer` and other repeatable character producers.
- `AbstractBinAsyncEntityProducer` clears its byte buffer, fixing repeatable binary subclasses.
- `DigestingEntityProducer` resets its `MessageDigest` before releasing its wrapped producer.

Regression tests cover file, path, string, repeatable binary, and digesting producers. Before the buffer fixes, replaying `abcdef` after only `abc` had been accepted produced `defabcdef`; after the fixes it produces exactly `abcdef`. Before the digest fix, the replay body was correct but its MD5 was calculated over `abcabcdef` instead of `abcdef`.

The same affected code is present on the `5.4.x` branch, so this change is also a candidate for backporting.

Tests:

- `./mvnw -pl httpcore5 -Dtest=TestStringAsyncEntityProducer,TestAbstractBinAsyncEntityProducer,TestFileAsyncEntityProducer,TestPathAsyncEntityProducer test` — 13 tests passed
- `./mvnw -pl httpcore5 -Dtest=TestDigestingEntityProducer test` — 2 tests passed
- `./mvnw -pl httpcore5 test` — 1,415 tests passed
- `./mvnw -pl httpcore5 verify -DskipTests` — checkstyle, Animal Sniffer, and japicmp passed


ok2c:
@mkurz That was quite sloppy on my part. Thank you for finding and reporting the defects.

ok2c:
Cherry-picked to `5.4.x`

PR #699, opened 2026-09-19
HTTPCORE-800: Fix hexadecimal encoding of control characters

arturobernalg:
Use the character value instead of its position in the message.


ok2c (approved):

RevathyRamalingam:
When is the release 5.4.4 expected @ok2c ?

